PR #5, opened 2020-06-02
ARM: tegra: Add device-tree for ASUS Transformer Prime TF201

clamor95:
This commit might not be merged yet, but is definitely cleaner then TF300T and TF700T device trees. Those transformer trees should be reworked.

If you decide to merge it anyway,  I will push one more patch after testing builds to fix all bugs I could find connected with tree.

Some parts of tf201 dts can be moved to common, but I wasn't 100% certain about their common nature among TFs and proper description of some stuff in split form.

### DIFF
--- a/arch/arm/boot/dts/tegra30-asus-tf201.dts
+++ b/arch/arm/boot/dts/tegra30-asus-tf201.dts
@@ -1,0 +1,718 @@
+// SPDX-License-Identifier: GPL-2.0
+/dts-v1/;
+
+/* This dts file describes the Asus Transformer Prime TF201 tablet */
+/* CPU Speedo ID 3, Soc Speedo ID 2, CPU Process: 1, Core Process: 0 */
+
+#include "tegra30-asus-transformer-common.dtsi"
+
+/ {
+	model = "Asus Transformer Prime TF201";
+	compatible = "asus,tf201", "nvidia,tegra30";
+
+	host1x@50000000 {
+		lcd: dc@54200000 {
+			rgb {
+				status = "okay";
+
+				port@0 {
+					lcd_output: endpoint {
+						remote-endpoint = <&lvds_encoder_input>;
+						bus-width = <24>;
+					};
+				};
+			};
+		};
+
+		hdmi: dc@54240000 { };
+
+		hdmi@54280000 {
+			status = "okay";
+
+			hdmi-supply = <&vdd_5v0_sys>;
+			pll-supply = <&vdd_1v8>;
+			vdd-supply = <&vdd_3v3_sys>;
+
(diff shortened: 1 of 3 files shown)


digetx:
It will be awesome if you could coordinate with @jja2000  and @osctobe in regards to the Transformers DTs clean up and unification. Will be great if you, guys, could create a minimal and clean DTs which could be upstreamed, the rest of the WIP features (like dock and EC drivers, etc) could be added in a separate commits, similarly to what we have now in a case of the Nexus 7 and A500 DTs.

Anyways, it's absolutely fine to carry and improve any of the patches in the grate-kernel. It's one of the purposes of the grate-kernel to brew experimental features until they are good for the upstream. I can easily drop patches, sqaush and rebase the grate-kernel tree in order apply new versions of the patches, so you can always open a new pull request with an updated version, although incremental updates should be a bit nicer to have for an easier review.

arch/arm/boot/dts/tegra30-asus-tf201.dts
```diff
@@ -0,0 +1,718 @@
...
+			max-speed = <4000000>;
+
+			clocks = <&tegra_pmc TEGRA_PMC_CLK_BLINK>;
+			clock-names = "txco";
+
+			vbat-supply  = <&vdd_3v3_sys>;
+			vddio-supply = <&vdd_1v8>;
+
+			device-wakeup-gpios = <&gpio TEGRA_GPIO(U, 1) GPIO_ACTIVE_HIGH>;	/* bt_ext_wake, out, lo	*/
+			host-wakeup-gpios =   <&gpio TEGRA_GPIO(U, 6) IRQ_TYPE_LEVEL_LOW>;	/* bt_host_wake, in, lo	*/
```
digetx, line 96: It's wrong to use `IRQ_TYPE_LEVEL_LOW` for the GPIO, it should be `GPIO_ACTIVE_LOW`. Please correct it in the update.

digetx:
The DT looks okay to me at a quick glance, nice work! :) Please keep up improving it, will be nice to have all the unnecessary comments removed from DT (or at least moved to a separate commit) and the DT formatting made similar to a Nexus 7 (to make it a bit more eye-candy), so that we could have a solid minimal base DT.

arch/arm/boot/dts/tegra30-asus-tf201.dts
clamor95: Thanks, I will correct this. It is a bit complicated for one person to track all code failures 😋

clamor95:
Currently we must have comments for tf's. It is really hard to find datasheets and all needed docs.

Three is made in same way as you made your. Good and clean work I can say. It inspired me A LOT! Common dtsi may be reused and more stuff can be moved there in next few commits during other tf trees reorganisation, so there will not be much of commit mess.

digetx:
Okay, looking forward to the next update :)

PR #10, opened 2020-10-01
Update README

pratTweaks:
improve docs


okias:
this file is inherited from https://git.kernel.org/pub/scm/linux/kernel/git/next/linux-next.git/

since it's not objective of this project change anything else than nvidia tegra and related code, if you want this change, you should try send it to linux-next. Anyway I don't see any improvement except adding quotes (not sure why).

For now closing

PR #11, opened 2020-11-08
Update common.c

tailormoon:
common.c fixed compilation error


tailormoon:
fixed compilation error

digetx:
What is the compilation error?

PR #14, opened 2020-12-24
Add device-tree for ASUS Transformer Prime TF201

clamor95:
@digetx we finally managed to reach grouper state of functionality on Transformer Prime. Works everything except sound and cameras (yet!). You can squash my tree commit with previous and remove wip status. Next commits for sound and camera will be applied as separate commits. Also Ion found issue with hotplug i2c (replug of dock failed), and proposed working patch. I'm asking to incorporate it in grate as well.

### DIFF
--- a/arch/arm/boot/dts/tegra30-asus-transformer-common.dtsi
+++ b/arch/arm/boot/dts/tegra30-asus-transformer-common.dtsi
@@ -59,10 +59,6 @@
 			ecc-size = <16>;
 		};
 
-//		lp0_vec@bddf9000 {
-//			reg = <0xbddf9000 0x2000>;	// passed from bootloader (ATAGS/NVIDIA, cmdline)
-//		};
-
 		trustzone@bfe00000 {
 			/* used by TF firmware */
 			reg = <0xbfe00000 0x200000>;	/* 2MB */
@@ -441,7 +437,7 @@
 				nvidia,function = "spi4";
 				nvidia,pull = <TEGRA_PIN_PULL_NONE>;
 				nvidia,tristate = <TEGRA_PIN_ENABLE>;
-				nvidia,enable-input = <TEGRA_PIN_ENABLE>;
+				nvidia,enable-input = <TEGRA_PIN_DISABLE>;
 			};
 			gmi_a18_pb1 {
 				nvidia,pins = "gmi_a18_pb1";
@@ -553,10 +549,16 @@
 			gmi_wp_n_pc7 {
 				nvidia,pins = "gmi_wp_n_pc7",
 						"gmi_wait_pi7",
-						"gmi_cs0_n_pj0",
-						"gmi_cs1_n_pj2",
-						"gmi_cs2_n_pk3",
 						"gmi_cs3_n_pk4";
+				nvidia,function = "rsvd1";
+				nvidia,pull = <TEGRA_PIN_PULL_NONE>;
+				nvidia,tristate = <TEGRA_PIN_ENABLE>;
+				nvidia,enable-input = <TEGRA_PIN_DISABLE>;
+			};
+			gmi_cs0_n_pj0 {
+				nvidia,pins = "gmi_cs0_n_pj0",
+						"gmi_cs1_n_pj2",
+						"gmi_cs2_n_pk3";
 				nvidia,function = "rsvd1";
 				nvidia,pull = <TEGRA_PIN_PULL_NONE>;
 				nvidia,tristate = <TEGRA_PIN_ENABLE>;
@@ -600,13 +602,6 @@
 				nvidia,enable-input = <TEGRA_PIN_ENABLE>;
 			};
 
-			pbb0 {
-				nvidia,pins = "pbb0";
-				nvidia,function = "rsvd2";
-				nvidia,pull = <TEGRA_PIN_PULL_NONE>;
-				nvidia,tristate = <TEGRA_PIN_DISABLE>;
-				nvidia,enable-input = <TEGRA_PIN_ENABLE>;
-			};
 			pcc1 {
 				nvidia,pins = "pcc1";
 				nvidia,function = "rsvd2";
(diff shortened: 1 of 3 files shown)


arch/arm/boot/dts/tegra30-asus-transformer-common.dtsi
```diff
@@ -1586,10 +1617,6 @@
 		status = "okay";
 	};
 
-	tegra-power-domain {
-		core-supply = <&vdd_core>;
-	};
-
```
digetx, line 1592: This probably shouldn't be removed?

clamor95: I have accidentally duplicated this binding. We fixed magnetometer so pull request needs to be updated.

clamor95:
@digetx this commits should be good to go. Review tree if it good enough to grate.

arch/arm/boot/dts/tegra30-asus-transformer-common.dtsi
```diff
@@ -1397,8 +1427,8 @@
 		clocks = <&tegra_pmc TEGRA_PMC_CLK_BLINK>;
 		clock-names = "ext_clock";
 
-		reset-gpios = <&gpio TEGRA_GPIO(D, 3) GPIO_ACTIVE_LOW>;	// wlan_rst, out, lo [wifi-off]
-		post-power-on-delay-ms = <300>;
+		reset-gpios = <&gpio TEGRA_GPIO(D, 3) GPIO_ACTIVE_HIGH>;
```
digetx, line 1430: Downstream kernel suggests that GPIO_ACTIVE_LOW was correct.

https://github.com/AndroidRoot/android_kernel_asus_tf201/blob/master/arch/arm/mach-tegra/board-cardhu-sdhci.c#L238

`cardhu_wifi_power()` sets both `PWR` and `RST` GPIOs to the same value, hence `RST` is active-low.

I assume WIFi doesn't work for you yet, correct?

arch/arm/boot/dts/tegra30-asus-transformer-common.dtsi
digetx: This is still not resolved after the update, but I can fixup it myself after the merge.

arch/arm/boot/dts/tegra30-asus-transformer-common.dtsi
```diff
@@ -1635,28 +1663,3 @@
...
-	/delete-node/ opp@750000000,1300;
-	/delete-node/ opp@800000000,1300;
-	/delete-node/ opp@900000000,1350;
-};
-
-&emc_bw_dfs_opp_table {
-	/delete-node/ opp@750000000;
-	/delete-node/ opp@800000000;
-	/delete-node/ opp@900000000;
-};
```
digetx, line 1649: This also shouldn't be removed?

clamor95: We don't know what nodes should be removed. Grouper and Ouya use different staff, hence of you can suggest what should be removed, no doubt, we will apply.

arch/arm/boot/dts/tegra30-asus-transformer-common.dtsi
clamor95: Other t30 devices doesn't have this binding, but have https://github.com/clamor95/linux/blob/master/arch/arm/boot/dts/tegra30-asus-transformer-common.dtsi#L1514-L1517 and I put both of them. If they both are needed, fix it please by yourself if you can.

arch/arm/boot/dts/tegra30-asus-transformer-common.dtsi
digetx: You should remove nodes that are not supported by the board in order not to expose them to drivers like devfreq, otherwise you'll see the unused rates in powertop and etc. This EMC change isn't critical, but it's better to do it properly.

If you're unsure about other transformers, then you could move it to the `tegra30-asus-tf201.dts`.

clamor95: I understand this, but where in board I can find what nodes aren't used?
I'd like to have this change, but since I don't understand where from values are taken, can't set them blindly.

arch/arm/boot/dts/tegra30-asus-transformer-common.dtsi
digetx: Aha, I missed to remove it from the transformers.dtsi by accident when was updating the nodes to `core-domain`. Hence this change is correct, thanks.

arch/arm/boot/dts/tegra30-asus-transformer-common.dtsi
digetx: You have memory timings defined in the `TF201.dts`, the max memory freq is 500MHz, hence everything above it is unsupported.

We should probably add the 500MHz entry to `tegra30-peripherals-opp.dtsi`, for consistency. I'll improve this myself after the merge.

clamor95: So you will add delete nodes as well? Would be perfect. Other TFs use other memory models (TF201 has LPDDR2 and other use LPDDR3) so placing it in tf201.dts would be clever.

arch/arm/boot/dts/tegra30-asus-transformer-common.dtsi
IonAgorria: We have WiFi and BT working with current state of DTS.

For what I can understand https://github.com/AndroidRoot/android_kernel_asus_tf201/blob/master/arch/arm/mach-tegra/board-cardhu-sdhci.c#L276 here is that both GPIO's are configured same way (active high due to absence of  GPIOF_ACTIVE_LOW)

I did tests with setting LOW and HIGH and seems that both work as long as D3 is cycled during bringup, so I consider ACTIVE_HIGH to be correct in reset, as enable only works in ACTIVE_HIGH too and both are configured same way in downstream.

arch/arm/boot/dts/tegra30-asus-transformer-common.dtsi
digetx: Yes, I'll add all the necessary nodes.

arch/arm/boot/dts/tegra30-asus-transformer-common.dtsi
digetx: The cardhu board file is used multiple different devices, it's possible that the PWR GPIO isn't connected to anything on TF201, i.e. power is fixed.

Could you please fill up this table:

| PWR (PD4)  | RST (PD3)  | WiFi works (yes/no)  |
| --------------- |:---------------:| -------------------------:|
| 1                  |  1                |                                  |
| 0                  |  0                |                                  |
| 1                  |  0                |                                  |
| 0                  |  1                |                                  |

IonAgorria: After doing tests, PWR must be HIGH or SDIO device is not present at all during dmesg

When brcm_wifi_pwrseq block is removed it doesn`t work, but if I remove PD3 gpio from pwrseq it still works as usual so seems that RST is not connected at all.

I also tried setting post-power-on-delay-ms in mmc controller without pwrseq but doesn't work.

We should remove reset-gpios only? the rest seems necessary as is.

digetx: It's probably exactly the same as Nexus 7 then, i.e. the PD4 is connected to the reset pin of the WiFi chip. Since hardware needs a delay after coming out from reset, you should set the reset-gpios of pwrseq to PD4 and specify power-gpio as fixed to 3.3v. You may copy it all from the nexus7.dtsi.

IonAgorria: After using Nexus 7 pwrseq it still works. I amended the changes to existing commit.

arch/arm/boot/dts/tegra30-asus-transformer-common.dtsi
```diff
@@ -1418,14 +1448,13 @@
 		non-removable;
 
 		mmc-pwrseq = <&brcm_wifi_pwrseq>;
-		vmmc-supply = <&vdd_3v3_sys>;
+		vmmc-supply = <&vdd_3v3_com>;
 		vqmmc-supply = <&vdd_1v8>;
 
 		wifi@1 {
 			reg = <1>;
-			compatible = "brcm,bcm4329-fmac";
```
digetx, line 1426: Why not to place this compatible in the transformer-common.dtsi?

clamor95: tf700t uses bcm4330

digetx: This compatible is universal for all BCM versions. Nexus 7 also has 4330, but compatible is the 4329.

clamor95: Only wifi works this way, or bt will as well?

digetx: Bluetooth has individual compatibles, but it should be okay to specify multiple compatibles in the common dtsi as well since the hardware description is exactly the same and driver takes care of probing the actual h/w version for firmware selection.

clamor95: Ok, we will try and let you know.

digetx: Alright, looks good.

PR #15, opened 2020-12-29
ARM: tegra: Add device-tree for ASUS Transformer PAD TF300T

clamor95:
We have refited previous TF300T device tree to work with common transformer device tree. To implement this commit I'd ask you to drop commits that I've reverted here from tree and add last commit in the sequence AFTER my commit for TF201 (since common tree was added there). We will add common transformers defconfig later on.

Additionally I'd ask you to amend and add https://github.com/grate-driver/linux/commit/e429ee243d95c0ef39e3ab661818b8ecc8b2f37d#diff-4a31af8261f8368ab91e9d684f2664249db2cf30d499f7d439fa902c8714d04fL1255-L1256, or you can add definition of dts in tf300t device tree commit directly.

Authorship of tf300t device tree commit may be changed to Michał Mirosław or left as is.

TF300T has same state of working staff as TF201 now. We have no TF700T dev/user that could test tree yet.

P. S. You may remove transformer branch of grate kernel since we have full support under master branch


digetx:
Well done, thank you!

PR #18, opened 2021-01-27
WIP: ARM: tegra: chagall fixes

cunidev:
Fix some minor misconfiguration in Pegatron Chagall pinmux

### DIFF
--- a/arch/arm/boot/dts/tegra30-pegatron-chagall.dts
+++ b/arch/arm/boot/dts/tegra30-pegatron-chagall.dts
@@ -446,7 +446,7 @@
 			};
 			spi2_sck_px2 {
 				nvidia,pins = "spi2_sck_px2";
-				nvidia,function = "spi2";
+				nvidia,function = "gmi";
 				nvidia,pull = <TEGRA_PIN_PULL_NONE>;
 				nvidia,tristate = <TEGRA_PIN_DISABLE>;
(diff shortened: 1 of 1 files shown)


digetx (commented):
Does this mean that Chagall don't have a USB host port? Internet suggests that Chagall only supports USB OTG, is this correct?

cunidev:
Yes. The usb3 bus may be supported through external docks though, which I am not in possession of? 
There is also an usb2 bus that is used by the 3G modem, but this is yet to be enabled due to the need of some specific enable pins.

cunidev:
The strange thing is, enabling usb3 apparently breaks usb1 for some reason.

digetx:
If USB3 physically presents in hardware, but not usable at the moment, then what about to set its `status` to "disabled" and add a comment saying `/* needs a dock station support */` instead of removing the nodes entirely?

cunidev:
Yes. There is also another issue, which I am currently investigating, on the USB bus. I added the WIP tag, please keep on hold for now.

digetx:
Please try to boot with `usbcore.autosuspend=-1` in kernel cmdline and see if it changes anything.

cunidev:
I tried looking deeper, and this MR seems to make more damage than everything so far. I'll close it, and re-open it when I reach a conclusion.

PR #22, opened 2021-03-09
Transformers and Chagall updates

clamor95:
I have included some changes into TF201, TF300T and Chagall device trees. Additionally I've changed commits to co-author system, since this would be the best option. I'd ask you to cherry-pick them instead of current ones in next master branch update. Thanks.


digetx:
"Co-authored-by" isn't a valid upstream tag, "Co-developed-by" is valid. Otherwise look good at a quick glance.

https://www.kernel.org/doc/html/latest/process/submitting-patches.html#when-to-use-acked-by-cc-and-co-developed-by

clamor95:
"Co-developed-by" does not add few author links as "Co-authored-by"

digetx:
The ""Co-developed-by"" require the "Signed-off-by" tag from each developer.

@osctobe, @cunidev are you okay if we will add yours "Signed-off-by" to the patches?

cunidev:
Yes, sure

clamor95:
Still looks silly

digetx:
@cunidev thanks

The Co-developed-by was correct, I'll re-add after merge. 

@osctobe I'll add yours s-b to the patch, please let me know if you have objections.

PR #28, opened 2021-04-08
enable DDR50 for tegra30 in nvquirks of sdhci driver

KaiJan57:
It works on my device at least…


digetx:
@pgwipeout Hi, could you please remind me whether you tried this MMC driver change on Ouya?

clamor95:
@digetx is this change relevant? Does it actually do smth? Does tegra30 supports this?

pgwipeout:
> @pgwipeout Hi, could you please remind me whether you tried this MMC driver change on Ouya?

Yes I did test DDR on Ouya. 
I had to subsequently reduce the clock rate further to maintain stability and such the performance was on par with SDR.

digetx:
@clamor95 Yes

On Nexus 7 EMMC before:

```
# cat /sys/kernel/debug/mmc0/ios 
clock:          52000000 Hz
actual clock:   51000000 Hz
vdd:            19 (3.1 ~ 3.2 V)
bus mode:       2 (push-pull)
chip select:    0 (don't care)
power mode:     2 (on)
bus width:      3 (8 bits)
timing spec:    1 (mmc high-speed)
signal voltage: 1 (1.80 V)
driver type:    0 (driver type B
```

after:

```
# cat /sys/kernel/debug/mmc0/ios 
clock:          52000000 Hz
actual clock:   52000000 Hz
vdd:            19 (3.1 ~ 3.2 V)
bus mode:       2 (push-pull)
chip select:    0 (don't care)
power mode:     2 (on)
bus width:      3 (8 bits)
timing spec:    8 (mmc DDR52)
signal voltage: 1 (1.80 V)
driver type:    0 (driver type B
```

@pgwipeout Thank you!

@KaiJan57  We could support DDR mode, but it should be made opt-in and MMC core should allow us to do that easily. I'll try to implement the opt-in support on top of this patch and then will merge this patch.

digetx:
@KaiJan57 Have you tried to benchmark the IO perf? What numbers have you got with and without the DDR mode?

KaiJan57:
Not so far, I just thought that it would be alright if downstream is allowing it as well?

digetx:
Not always.

maximschwalm:
Wouldn't adding the `mmc-ddr-1_8v` property be enough? Like here: https://github.com/grate-driver/linux/commit/dc741b7bde05a7d1e4382c7c633667fbda0f27cc

KaiJan57:
> Wouldn't adding the `mmc-ddr-1_8v` property be enough? Like here: [dc741b7](https://github.com/grate-driver/linux/commit/dc741b7bde05a7d1e4382c7c633667fbda0f27cc)

I don't think that would be enough, as I can't see how this:
https://github.com/grate-driver/linux/blob/d43dd42b07aa97634fb5399819157af972f92835/drivers/mmc/host/sdhci-tegra.c#L406-L407
can be triggered just by using a device-tree property… But correct me if I am mistaken.

digetx:
The `SDHCI_MISC_CTRL_ENABLE_DDR50` bit isn't documented in the T30 TRM. This means that DDR50 mode should be exposed by default and quirk is not needed on T30.

KaiJan57:
Well, then I agree on the device-tree entry being the best solution, thank you very much!

KaiJan57:
So I just wanted to give the recent changes a try but compilation of device trees fails with `TypeError: __init__() got an unexpected keyword argument 'resolver'`. Any ideas what I am doing wrong?

digetx:
This is a python error. I haven't encountered that problem, apparently it should come from some build script. Need more context.

PR #30, opened 2021-04-14
ARM: tegra: Lenovo IdeaTab A2109A:  enable charging with closely rela…

KaiJan57:
…ted smb345 driver, add backlight brightness calibration data, add thermal-zones

### DIFF
--- a/arch/arm/boot/dts/tegra30-lenovo-ideatab-a2109a.dts
+++ b/arch/arm/boot/dts/tegra30-lenovo-ideatab-a2109a.dts
@@ -4,6 +4,7 @@
 #include <dt-bindings/input/gpio-keys.h>
 #include <dt-bindings/input/input.h>
 #include <dt-bindings/mfd/max77620.h>
+#include <dt-bindings/power/summit,smb347-charger.h>
 
 #include "tegra30.dtsi"
 #include "tegra30-cpu-opp.dtsi"
@@ -25,6 +26,15 @@
 		serial1 = &uartb; /* GPS */
 		serial2 = &uartc; /* Bluetooth */
 	};
+
+    battery: battery-cell {
+        compatible = "simple-battery";
+        constant-charge-current-max-microamp = <2000000>;
+        operating-range-celsius = <0 50>;
+        alert-celsius = <3 42>;
+
+		voltage-min-design-microvolt = <>;
+    };
 
 	/*
 	 * The decompressor and also some bootloaders rely on a
@@ -1085,6 +1095,7 @@
 		fuelgauge: bq27541@55 {
 			compatible = "ti,bq27541";
 			reg = <0x55>;
+			monitored-battery = <&battery>;
 		};
 
 		pmic: pmic@3c {
@@ -1291,6 +1302,18 @@
 				};
 			};
 		};
+
+		power_supply: charger@1b {
+			compatible = "summit,smb345"; // actually is closely related smb349, but this makes charging work
+			reg = <0x1b>;
+
+			summit,usb-current-limit-microamp = <500000>;
+			summit,enable-charge-control = <SMB3XX_CHG_ENABLE_SW>;
+			summit,enable-usb-charging;
+			summit,enable-otg-charging;
+
+			monitored-battery = <&battery>;
+		};
 	};
 
 	pmc@7000e400 {
@@ -1351,7 +1374,6 @@
 		vmmc-supply = <&vddio_sdmmc>;
 		vqmmc-supply = <&vdd_1v8>;
 
-		/* Azurewave AW-NH665 BCM4330 */
 		wifi@1 {
 			reg = <1>;
 			compatible = "brcm,bcm4329-fmac";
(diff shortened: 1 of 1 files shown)


arch/arm/boot/dts/tegra30-lenovo-ideatab-a2109a.dts
```diff
@@ -1354,7 +1377,7 @@
 		/* Azurewave AW-NH665 BCM4330 */
 		wifi@1 {
 			reg = <1>;
-			compatible = "brcm,bcm4329-fmac";
+			compatible = "brcm,bcm4330-fmac";
```
digetx, line 1380: This is a wrong compatible, old was correct.

arch/arm/boot/dts/tegra30-lenovo-ideatab-a2109a.dts
```diff
@@ -1393,9 +1416,41 @@
...
+			185 185 186 187 187 188 189 189
+			190 191 192 193 194 195 196 197
+			198 199 200 201 201 202 203 203
+			204 205 205 206 207 207 208 209
+			209 210 211 211 212 212 213 213
+			214 215 215 216 216 217 217 218
+			219 219 220 222 226 230 232 234
+			236 238 240 244 248 251 253 255
+		>;
+		default-brightness-level = <10>;
```
digetx, line 1453: I made commit that reverts backlight, but forgot to push it. It's good that you reminded about it, I'll drop my patch.

arch/arm/boot/dts/tegra30-lenovo-ideatab-a2109a.dts
```diff
@@ -1085,6 +1095,7 @@
 		fuelgauge: bq27541@55 {
 			compatible = "ti,bq27541";
 			reg = <0x55>;
+			monitored-battery = <&battery>;
```
digetx, line 1098: I'm now thinking that it should be better to remove the monitored-battery node from Nexus7 DT because battery monitor kernel driver will try to reprogram controller firmware based on the values from the battery node, it fails to do so because controller was locked by manufacturer for Nexus7 (there is a kernel message about it), but I'm not sure that this is the case for all other devices.

arch/arm/boot/dts/tegra30-lenovo-ideatab-a2109a.dts
```diff
@@ -1578,6 +1633,47 @@
...
+				map0 {
+					trip = <&trip0>;
+					cooling-device = <&cpu0 THERMAL_NO_LIMIT THERMAL_NO_LIMIT>,
+							 <&cpu1 THERMAL_NO_LIMIT THERMAL_NO_LIMIT>,
+							 <&cpu2 THERMAL_NO_LIMIT THERMAL_NO_LIMIT>,
+							 <&cpu3 THERMAL_NO_LIMIT THERMAL_NO_LIMIT>;
+				};
+			};
+		};
+	};
```
digetx, line 1675: In the latest version of the sensor patch series I moved the thermal zone to tegra30.dtsi and it's now universally available on all devices, hence there is no need to add it to the lenovo DT.

arch/arm/boot/dts/tegra30-lenovo-ideatab-a2109a.dts
digetx: To be more correct, I meant the monitored-battery `property` instead of the `node`. The battery-cell node itself should be okay to add.

arch/arm/boot/dts/tegra30-lenovo-ideatab-a2109a.dts
KaiJan57: Well, both work so it obviously does not matter, but `/* Azurewave AW-NH665 BCM4330 */` is actually a leftover from what I had copied, so – correct me if I am wrong – I thought it would be less irritating to just put the right chip name in place…

arch/arm/boot/dts/tegra30-lenovo-ideatab-a2109a.dts
KaiJan57: I can't see any dmesg error messages related to `fuelgauge` on cl2n boot, so that problem is probably specific to Nexus7. But as with many of the more recent changes to this device-tree: it does not really make a big difference.

digetx: Do have `bq27xxx-battery 0-0055: data memory update not supported for chip` message in kmsg?

arch/arm/boot/dts/tegra30-lenovo-ideatab-a2109a.dts
digetx: Upstream kernel requires device trees to conform to the device tree binding specification (please see `https://github.com/grate-driver/linux/blob/master/Documentation/devicetree/bindings/net/wireless/brcm,bcm4329-fmac.yaml'). It happens to work because brcmfmac driver also matches device by the MMC card ID and in this case the optional interrupt won't be used by driver because device-tree is skipped.

If you want to explicitly specify that this is a bcm4330, then the correct variant should be `compatible = "brcm,bcm4329-fmac", "brcm,bcm4330-fmac"`.

arch/arm/boot/dts/tegra30-lenovo-ideatab-a2109a.dts
KaiJan57: I'm sorry, I do. Did not conceive that as an error…

PR #31, opened 2021-04-14
ARM: tegra: Lenovo IdeaTab A2109A: fixed PR

KaiJan57:
ARM: tegra: Lenovo IdeaTab A2109A: enable charging with closely related smb345 driver, add backlight brightness calibration data, temporarily change volume-up to power-key, as sound does not work anyway

### DIFF
--- a/arch/arm/boot/dts/tegra30-lenovo-ideatab-a2109a.dts
+++ b/arch/arm/boot/dts/tegra30-lenovo-ideatab-a2109a.dts
@@ -4,6 +4,7 @@
 #include <dt-bindings/input/gpio-keys.h>
 #include <dt-bindings/input/input.h>
 #include <dt-bindings/mfd/max77620.h>
+#include <dt-bindings/power/summit,smb347-charger.h>
 
 #include "tegra30.dtsi"
 #include "tegra30-cpu-opp.dtsi"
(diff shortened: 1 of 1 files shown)


arch/arm/boot/dts/tegra30-lenovo-ideatab-a2109a.dts
```diff
@@ -1291,6 +1302,18 @@
...
+			compatible = "summit,smb345"; // actually is closely related smb349, but this makes charging work
+			reg = <0x1b>;
+
+			summit,usb-current-limit-microamp = <500000>;
+			summit,enable-charge-control = <SMB3XX_CHG_ENABLE_SW>;
+			summit,enable-usb-charging;
+			summit,enable-otg-charging;
+
+			monitored-battery = <&battery>;
+		};
```
digetx, line 1316: It's okay to use the smb345 compatible for now, but we will need to add the proper smb349 support. I see that downstream tegra and rockchip kernels have a custom driver for smb349, so we could borrow at least some smb349 h/w parameters from them.

digetx:
It shouldn't be difficult to support the power button properly, I'll try to take a closer look at soon. Thanks!

PR #32, opened 2021-04-15
ARM: tegra: lenovo-a2109a: small fixes

KaiJan57:
ARM: tegra: lenovo-a2109a: remove fuelgauge battery node, increase charger usb-current-limit, add wifi/bt module product name based on firmware signature

### DIFF
--- a/arch/arm/boot/dts/tegra30-lenovo-ideatab-a2109a.dts
+++ b/arch/arm/boot/dts/tegra30-lenovo-ideatab-a2109a.dts
@@ -1095,7 +1095,6 @@
 		fuelgauge: bq27541@55 {
 			compatible = "ti,bq27541";
 			reg = <0x55>;
-			monitored-battery = <&battery>;
 		};
 
 		pmic: pmic@3c {
@@ -1307,7 +1306,6 @@
 			compatible = "summit,smb345"; // actually is closely related smb349, but this makes charging work
 			reg = <0x1b>;
 
-			summit,usb-current-limit-microamp = <500000>;
 			summit,enable-charge-control = <SMB3XX_CHG_ENABLE_SW>;
 			summit,enable-usb-charging;
 			summit,enable-otg-charging;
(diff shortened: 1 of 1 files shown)


arch/arm/boot/dts/tegra30-lenovo-ideatab-a2109a.dts
```diff
@@ -1307,7 +1306,7 @@
 			compatible = "summit,smb345"; // actually is closely related smb349, but this makes charging work
 			reg = <0x1b>;
 
-			summit,usb-current-limit-microamp = <500000>;
+			summit,usb-current-limit-microamp = <1500000>;
```
digetx, line 1309: Have you checked that the programmed register value is okay for smb349?

okias:
I wanted make a comment for previous PR, but never too late. Check downstream smb349 driver, try compare values in arrays in mainline smb347-charger if matches with smb345 probably leave it as it's, but if there are differences, send patch and add "smb349" variant.

arch/arm/boot/dts/tegra30-lenovo-ideatab-a2109a.dts
KaiJan57: https://github.com/Unlegacy-Android/android_kernel_nvidia_tegra3/blob/ab3502c38c3fb2e9410212497c0cb300935932c0/arch/arm/mach-tegra/board-cl2n.c#L218
According to downstream kernel, smb349 can even handle up to 2 A!

digetx: You have set the compatibe to sm345, but if register isn't compatible with the smb349, then you program something else.

Also, looking at  https://github.com/Unlegacy-Android/android_kernel_nvidia_tegra3/blob/ab3502c38c3fb2e9410212497c0cb300935932c0/drivers/power/smb349-charger.c I see that the code skips the max-current programming at all, hence the board value has no effect.

digetx: The smb349 registers also should be non-volatile. Hence once you programmed invalid value, it will stick until you'll re-program it with the correct one, i.e. reboot and battery re-plug won't help to recover.

Let's remove the `usb-current-limit-microamp` from the DT until we will know that it works correctly.

KaiJan57: You have a point, I will change it.

digetx:
Looks good, thank you!

PR #33, opened 2021-04-16
ARM: tegra: lenovo-a2109a: Fix volume down and up keys switched

KaiJan57:


digetx:
Thanks!

KaiJan57:
Is there any chance to get HDMI audio working without codec driver, btw?

digetx:
No, only codec method is supported by the upstream kernel. Do you have any problems with it?

KaiJan57:
I hoped that sound over HDMI was not related to the codec IC, but as it is, there unfortunately is no sound even when using HDMI.

digetx:
I'm recalling that @pgwipeout was saying that HDMI codec needs to be explicitly whitelisted for pulseaudio, maybe this is the problem.

KaiJan57:
Interesting, do you know how, or could you point me to a guide?

digetx:
I don't, but maybe @pgwipeout could.

pgwipeout:
Apologies for the late response @KaiJan57.
You have to use alsamixer and make sure the hdmi sink is unmuted.
Use `speaker-test -Dhw:0,3 -c 2` to send a test audio stream to the device.
For pulseaudio add `load-module module-alsa-sink device=hw:0,3` to `/etc/pulse/default.pa`.

KaiJan57:
Thank you very much, I will give that a try soon. No need to apologize :)

KaiJan57:
Does not seem to work on my device, alsamixer returns:
Mixer attach default error: No such file or directory
Am I missing some kernel module maybe?

pgwipeout:
Ensure CONFIG_SND_HDA_TEGRA is either built in or loaded as a module.
Also ensure it's enabled in your device tree.

KaiJan57:
Yep, hda was missing in device tree. Now its working like a charm, thank you very much!

digetx:
Good to know that the HDMI audio problem is resolved, I assume that PR is incoming :)

PR #34, opened 2021-04-20
Asus Transformers update

clamor95:
Common:
- exclude dc node to common;
- update hog name and remove comment;
- set vio to 1.8v;
- add pmic gpios to fixed regs;
- set pwms in backlight as in downstream;
- update thermal;

TF300T:
- add init hog gpio;
- fix sound routing;

Squash please with propper commits. Thanks!


digetx:
Looks good, thank you!

PR #36, opened 2021-04-22
Some sdhci changes

clamor95:
Those are Michał Mirosław changes, but I've tested them for a long time already.

Not sure if all of changes are relevant for us. Thanks.


clamor95:
Yeah, sure. Btw, about serial. Tegra114 still doesn't have support in that driver. It is just skipped.

And this one https://github.com/grate-driver/linux/pull/36/commits/f67a3eb2c06b0ae8b0f23169f61c2895d6b67082 may be considerable, since as I know all t20 and t30 transformers have.

`mmcX: Invalid maximum block size, assuming 512 bytes`

digetx:
Tegra TRM says that only 512b is supported. I'm also not sure whether larger block sizes could make a positive difference in practice.

PR #39, opened 2021-06-04
Update Asus Transformers

clamor-s:
- since emmc needs 3.3v to work anyway. Set DDR52 mode
- gpio keys are splitted into buttons (power, vol up and down) and extcon keys (lid, ac, line out)
- sound card names changes to use common ucm (I'm 100% sure rt5631 is exactly same on all transformers that use it; I'm quite sure wm8903 is same on tf300t, chagall and tf101 but that needs proofs)
- changed labeling backlight from backlight_lvds to simple backlight
- changed regulators labeling. Added pmic hog gpio based on Michal's device tree. Changed fixed regulator approach. Regulator 0 is ac_bat 5v, Regulator 1-4 have pmic grios in, Regulator 5,6 are panel regulators, Regulator 7-10 are other fixed regulators neede for proper work.

Everything except emmc works fine. Tested by me, Max, Ihor and Andy.

P. S. May you add tested by to every tf tree? I'll provide lists. 
TF201 (me and Ion) so I assume none

TF300T
Tested-by: Ihor Didenko <tailormoon@rambler.ru>
Tested-by: Andreas Westman Dorcsak <hedmoo@yahoo.com>

TF700T
Tested-by: Maxim Schwalm <maxim.schwalm@gmail.com>
Tested-by: Andreas Westman Dorcsak <hedmoo@yahoo.com>
Tested-by: Jasper Korten <jja2000@gmail.com>

### DIFF
--- a/arch/arm/boot/dts/tegra30-asus-tf201.dts
+++ b/arch/arm/boot/dts/tegra30-asus-tf201.dts
@@ -137,7 +137,7 @@
 		compatible = "panel-lvds";
 
 		power-supply = <&vdd_pnl>;
-		backlight = <&backlight_lvds>;
+		backlight = <&backlight>;
 
 		width-mm = <217>;
 		height-mm = <136>;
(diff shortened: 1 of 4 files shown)


digetx:
I'll add the t-b's, thank you.

PR #42, opened 2021-06-26
Add support for Microsoft Surface RT

jenneron:
Still needs some testing. When it is done i will squash commits except one which adds efi dts.

### DIFF
--- a/arch/arm/boot/dts/tegra30-microsoft-surface-rt.dts
+++ b/arch/arm/boot/dts/tegra30-microsoft-surface-rt.dts
@@ -1,0 +1,288 @@
+// SPDX-License-Identifier: GPL-2.0
+
+/dts-v1/;
+
+#include <dt-bindings/input/input.h>
+#include "tegra30.dtsi"
+
+/ {
+	model = "Microsoft Surface RT";
+	compatible = "microsoft,surface-rt", "nvidia,tegra30";
+
+	aliases {
+		mmc0 = &sdmmc4; /* eMMC */
+		mmc1 = &sdmmc1; /* uSD slot */
+		rtc0 = &pmic;
+		rtc1 = "/rtc@7000e000";
+		serial0 = &uarta;
+	};
+
+	chosen {
+		stdout-path = "serial0:115200n8";
+	};
+
+	/* 2GB DDR3 Memory */
+	memory@80000000 {
+		reg = <0x80000000 0x40000000>; // 1 GiB is wrong size but 2GB doesnt work atm
+	};
+
+	host1x@50000000 {
+		hdmi@54280000 {
+			status = "okay";
+			nvidia,ddc-i2c-bus = <&hdmi_ddc>;
+			nvidia,hpd-gpio = <&gpio TEGRA_GPIO(N, 7) GPIO_ACTIVE_HIGH>;
+		};
+
+		dc@54200000 {
+			rgb {
+				status = "okay";
+				nvidia,panel = <&panel>;
+
+				port@0 {
+					lcd_output: endpoint {
+						remote-endpoint = <&lvds_encoder_input>;
+						bus-width = <24>;
+					};
+				};
+			};
+		};
+	};
+
+	pinmux@70000868 {
+		pinctrl-names = "default";
+		pinctrl-0 = <&state_default>;
+
+		state_default: pinmux {
+			/* HDMI HOTPLUG_DETECT */
+			hdmi-int-pn7 {
+				nvidia,pins = "hdmi_int_pn7";
+				nvidia,function = "hdmi";
+				nvidia,pull = <TEGRA_PIN_PULL_NONE>;
+				nvidia,tristate = <TEGRA_PIN_ENABLE>;
+				nvidia,enable-input = <TEGRA_PIN_ENABLE>;
+			};
+
+			volKeys {
+				nvidia,pins = "kb_row14_ps6", //down
+							  "kb_row15_ps7"; //up
+				nvidia,pull = <TEGRA_PIN_PULL_NONE>;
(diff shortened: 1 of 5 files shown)


digetx:
Is it ready?

jenneron:
> Is it ready?

Nope, PR is draft because of that. The first commit authored by CrackTheSurface works, i added 2 more commits for clean up, but currently it has problems with eMMC and SD card for some reason. It needs to be tested without those aliases. I don't know how it could affect, but i didn't change anything else related to eMMC and SD.

Also APX + uboot variant is not tested after my changes.

I will mark PR as ready when it is ready.

digetx:
There were some new errors related to MMC aliases in yesterday's linux-next, maybe it's related. I haven't looked at it closely yet.

jenneron:
Well, SDHCI is initialized, SD and eMMC are not detected.
https://dpaste.com/CJ7HL52ZH

arch/arm/boot/dts/tegra30-microsoft-surfacert.dts
```diff
@@ -0,0 +1,288 @@
+// SPDX-License-Identifier: GPL-2.0
+
+/dts-v1/;
+
+#include <dt-bindings/input/input.h>
+#include "tegra30.dtsi"
+
+/ {
+	model = "Microsoft Surface RT";
+	compatible = "microsoft,surfacert", "nvidia,tegra30";
```
digetx, line 10: Do we have only a single hardware variant of the SurfaceRT tablet?

Nit: what about "microsoft,surface-rt"?

arch/arm/boot/dts/tegra30-microsoft-surfacert-efi.dts
```diff
@@ -0,0 +1,22 @@
...
+		regulator-name = "vddio_sdmmc";
+		regulator-min-microvolt = <3300000>;
+		regulator-max-microvolt = <3300000>;
+		regulator-always-on;
+	};
+};
+
+&pmic {
+	/delete-node/ regulators;
+};
```
digetx, line 22: We could also override the board's compatible like this:

```
	compatible = "microsoft,surface-rt-efi",  "microsoft,surface-rt", "nvidia,tegra30";
```

Then we could replace CONFIG_ARM_ACPI_PARKING_PROTOCOL with:

```patch
diff --git a/arch/arm/include/asm/acpi_parking_protocol.h b/arch/arm/include/asm/acpi_parking_protocol.h
index aa5e6ba7918d..cce913687bbb 100644
--- a/arch/arm/include/asm/acpi_parking_protocol.h
+++ b/arch/arm/include/asm/acpi_parking_protocol.h
@@ -7,7 +7,7 @@ extern const struct smp_operations acpi_parking_protocol_ops;
 
 static inline bool acpi_parking_protocol_available(void)
 {
-	return IS_ENABLED(CONFIG_ARM_ACPI_PARKING_PROTOCOL);
+	return of_machine_is_compatible("microsoft,surface-rt-efi");
 }
 
 #endif /* __ASM_ARM_ACPI_PARKING_PROTOCOL_H */

```

arch/arm/boot/dts/tegra30-microsoft-surfacert.dts
jenneron: There are 2 T114-based surface tablets, but i haven't heard about other variants of T30-based Surface RT.

I assume i need to change it in the name of dts file too?

arch/arm/boot/dts/tegra30-microsoft-surfacert-efi.dts
jenneron: Ok, pushed, but not tested yet

arch/arm/boot/dts/tegra30-microsoft-surface-rt.dts
```diff
@@ -0,0 +1,288 @@
...
+		hdmi@54280000 {
+			status = "okay";
+			nvidia,ddc-i2c-bus = <&hdmi_ddc>;
+			nvidia,hpd-gpio = <&gpio TEGRA_GPIO(N, 7) GPIO_ACTIVE_HIGH>;
+		};
+
+		dc@54200000 {
+			rgb {
+				status = "okay";
+				nvidia,panel = <&panel>;
```
digetx, line 39: Why do you need legacy `nvidia,panel` if you have graph? You have a warning in KMSG because of it.

arch/arm/boot/dts/tegra30-microsoft-surface-rt.dts
```diff
@@ -0,0 +1,288 @@
...
+			panel_input: endpoint {
+				remote-endpoint = <&lvds_encoder_output>;
+			};
+		};
+	};
+
+	gpio-keys {
+		compatible = "gpio-keys";
+
+		label = "GPIO Buttons";
```
digetx, line 218: Does this label have any real effect? The gpio-keys binding suggests it's invalid.

You can check the device tree by:

```
# pip3 install --user dtschema
# make dtbs_check
```
I can fix it locally.

arch/arm/boot/dts/tegra30-microsoft-surfacert-efi.dts
digetx: It should also include `linux/of.h`, but I'll fix it locally.

arch/arm/boot/dts/tegra30-microsoft-surface-rt.dts
digetx: I'll fix it locally.

digetx:
Overall looks good, I'll correct the minor problems after the merge. Thank you!

PR #43, opened 2021-06-28
Update asus-ec driver

clamor-s:
d42633f and dts mains binding may be removed they only provide info about AC plug events. Deside by yourself.

P. S. Sound still does not work on transformers. Those with rt5631 have working ucm and booting from kexec enables sound. TF300T with wm8903 is in worst state. I'm starting think that I misbinded smth.


digetx:
Thanks!

PR #44, opened 2021-06-29
Asus EC yaml fixes and new dock revision

clamor-s:
Yaml I tried to fix. Should be fine. Additionally, we discovered new t30 transformer dock revision.


digetx:
Thanks!

PR #45, opened 2021-07-03
WIP: ARM: tegra: Support for Surface RT Type/Touch/Power-Cover; Display fix

CrackTheSurface:
# Type/Touch/Power-Cover Support for Surface RT:
tCover is connected via HID over I2C.
Sadly the slave address is 0x00 which is reserved as general call address.
The patch allows the I2C driver only to accept 0x00 as valid address if the device is a Surface RT.
`i2c-hotplug-gpio` allows to hotplug the tCovers.
## Tests
Tested on Surface RT with latest grate-linux kernel.
Plugged in the first cover after boot and changed the covers randomly.

Tested with following covers:
- TypeCover 1.Gen
- TypeCover 2.Gen
- TouchCover 1.Gen
- TouchCover 2.Gen

VendorID and ProductID were read correctly and every cover worked after hotplug.

# Display fix:
Added the property: `nvidia,panel`

### DIFF
--- a/arch/arm/boot/dts/tegra30-microsoft-surface-rt.dts
+++ b/arch/arm/boot/dts/tegra30-microsoft-surface-rt.dts
@@ -36,6 +36,8 @@
 		dc@54200000 {
 			rgb {
 				status = "okay";
+				
+				nvidia,panel = <&panel>;
 
 				port@0 {
 					lcd_output: endpoint {
--- a/arch/arm/boot/dts/tegra30-microsoft-surface-rt.dts
+++ b/arch/arm/boot/dts/tegra30-microsoft-surface-rt.dts
@@ -89,6 +89,31 @@
 	pwm@7000a000 {
 		status = "okay";
 	};
+	
+	i2c1: i2c@7000c000 {
+		status = "okay";
+		clock-frequency = <400000>;
+	};
+	
+	cover-i2c {
+		compatible = "i2c-hotplug-gpio";
+		
+		#address-cells = <1>;
+		#size-cells = <0>;
+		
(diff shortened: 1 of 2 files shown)


arch/arm/boot/dts/tegra30-microsoft-surface-rt.dts
```diff
@@ -36,6 +36,8 @@
 		dc@54200000 {
 			rgb {
 				status = "okay";
+				
+				nvidia,panel = <&panel>;
```
digetx, line 40: The root of the problem should be in the bridge node.

@jenneron confirmed on IRC that this change works:

```patch
diff --git a/arch/arm/boot/dts/tegra30-microsoft-surface-rt.dts b/arch/arm/boot/dts/tegra30-microsoft-surface-rt.dts
index f386cae6bc03..1c27d8a43ac9 100644
--- a/arch/arm/boot/dts/tegra30-microsoft-surface-rt.dts
+++ b/arch/arm/boot/dts/tegra30-microsoft-surface-rt.dts
@@ -248,7 +248,7 @@ volume-up {
        lvds-encoder {
                compatible = "lvds-encoder";
 
-               powerdown-gpios = <&gpio TEGRA_GPIO(B, 2) GPIO_ACTIVE_HIGH>;
+               powerdown-gpios = <&gpio TEGRA_GPIO(B, 2) GPIO_ACTIVE_LOW>;
 
                ports {
                        #address-cells = <1>;
```

Could you please test it?

CrackTheSurface: Works as intended :+1:

digetx: I'll merge this PR, then drop the nvidia,panel and fix the bridge node,

digetx:
Thank you!

PR #46, opened 2021-07-05
WIP: ARM: dts: tegra30-microsoft-surfacert: fix gpio-keys

jenneron:
Sorry for sending so small update, we missed this while testing first PR.
`KEY_LEFTMETA` and `KEY_POWER` had default state 1 in evtest.


digetx:
Small updates are actually better since it's easier to review them. Thank you!

PR #48, opened 2021-07-05
Asus Transformers fixes 

maximschwalm:
- Fix booting from recovery
- Fix RT5631 registers being out of sync after resuming from suspend
- Fix backlight supply
- Add some missing supplies
- Reduce W1 ramp delay to 20ms
- Choose better regulator names

### DIFF
--- a/arch/arm/boot/dts/tegra30-asus-transformer-common.dtsi
+++ b/arch/arm/boot/dts/tegra30-asus-transformer-common.dtsi
@@ -1221,7 +1221,7 @@
 
 			regulators {
 				vdd1 {
-					regulator-name = "vddio_ddr_1v2";
+					regulator-name = "vdd_1v2_gen";
 					regulator-min-microvolt = <600000>;
 					regulator-max-microvolt = <1500000>;
 					regulator-always-on;
@@ -1230,7 +1230,7 @@
 				};
 
 				vdd2_reg: vdd2 {
-					regulator-name = "vdd_1v5_gen";
+					regulator-name = "vddio_ddr";
 					regulator-min-microvolt = <600000>;
 					regulator-max-microvolt = <1500000>;
 					regulator-always-on;
@@ -1261,7 +1261,7 @@
 
 				/* eMMC VDD */
 				vcore_emmc: ldo1 {
-					regulator-name = "vdd_pexa,vdd_pexb";
+					regulator-name = "vdd_emmc";
 					regulator-min-microvolt = <1000000>;
 					regulator-max-microvolt = <3300000>;
 					regulator-always-on;
@@ -1269,9 +1269,9 @@
 
 				/* uSD slot VDD */
 				vdd_usd: ldo2 {
-					regulator-name = "vdd_sata,avdd_plle";
-					regulator-min-microvolt = <2900000>;
-					regulator-max-microvolt = <2940000>;
+					regulator-name = "vdd_usd";
+					regulator-min-microvolt = <3100000>;
+					regulator-max-microvolt = <3100000>;
 					regulator-always-on;
 				};
 
@@ -1279,7 +1279,7 @@
 				vddio_usd: ldo3 {
 					regulator-name = "vddio_usd";
 					regulator-min-microvolt = <1800000>;
-					regulator-max-microvolt = <3300000>;
+					regulator-max-microvolt = <3100000>;
 				};
 
 				ldo4 {
@@ -1371,13 +1371,14 @@
 
 	vdd_1v5_ddr: regulator@3 {
 		compatible = "regulator-fixed";
-		regulator-name = "mem_vddio_ddr";
+		regulator-name = "vdd_ddr";
 		regulator-min-microvolt = <1500000>;
 		regulator-max-microvolt = <1500000>;
 		regulator-always-on;
 		regulator-boot-on;
 		gpio = <&pmic 7 GPIO_ACTIVE_HIGH>;
 		enable-active-high;
+		vin-supply = <&vdd_5v0_bat>;
 	};
 
 	vdd_3v3_sys: regulator@4 {
@@ -1389,6 +1390,7 @@
 		regulator-boot-on;
 		gpio = <&pmic 6 GPIO_ACTIVE_HIGH>;
 		enable-active-high;
+		vin-supply = <&vdd_5v0_bat>;
 	};
 
 	vdd_pnl: regulator@5 {
@@ -1396,7 +1398,7 @@
 		regulator-name = "vdd_panel";
 		regulator-min-microvolt = <3300000>;
 		regulator-max-microvolt = <3300000>;
-		regulator-enable-ramp-delay = <300000>;
+		regulator-enable-ramp-delay = <20000>;
 		gpio = <&gpio TEGRA_GPIO(W, 1) GPIO_ACTIVE_HIGH>;
 		enable-active-high;
 		vin-supply = <&vdd_3v3_sys>;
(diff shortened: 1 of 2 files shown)


arch/arm/boot/dts/tegra30-asus-transformer-common.dtsi
```diff
@@ -1261,31 +1261,31 @@
...
+					regulator-max-microvolt = <3100000>;
 					regulator-always-on;
 				};
 
 				/* uSD slot VDDIO */
 				vddio_usd: ldo3 {
 					regulator-name = "vddio_usd";
 					regulator-min-microvolt = <1800000>;
-					regulator-max-microvolt = <3300000>;
+					regulator-max-microvolt = <3100000>;
```
digetx, line 1282: Doesn't this result in a error from MMC core when it tries to set 3.3v mode? I think this should stay 3.3v.

arch/arm/boot/dts/tegra30-asus-transformer-common.dtsi
```diff
@@ -1221,18 +1221,18 @@
...
 					ti,regulator-ext-sleep-control = <8>;
 				};
 
 				vdd2_reg: vdd2 {
-					regulator-name = "vdd_1v5_gen";
-					regulator-min-microvolt = <600000>;
-					regulator-max-microvolt = <1500000>;
+					regulator-name = "vddio_ddr";
+					regulator-min-microvolt = <1200000>;
+					regulator-max-microvolt = <1200000>;
```
digetx, line 1235: Are you sure that these voltages are suitable for all transformers?

digetx:
What change fixed panel after rebooting?

arch/arm/boot/dts/tegra30-asus-transformer-common.dtsi
maximschwalm: I don't think that 3.3v will be used. The kernel tries to match the vmmc and vqmmc voltage. That's why I can't set vqmmc lower than what vmmc is set to, because then the kernel complains that the voltage would be out of range

maximschwalm:
> What change fixed panel after rebooting?

Removing the ramp delay

arch/arm/boot/dts/tegra30-asus-transformer-common.dtsi
maximschwalm: Couldn't check it they are the same for all Transformers. At least TF300T seems to use the same voltages: [https://rere.qmqm.pl/git/?p=linux;a=blob;f=arch/arm/boot/dts/tegra30-asus-tf300t.dts;hb=refs/heads/tf300t](url)

digetx:
Do you have any idea why ramp delay causes the problem?

maximschwalm:
> Do you have any idea why ramp delay causes the problem?

I don't have any explanation. I discovered this purely by trial and error

digetx:
The original downstream kernel seems use 20ms delay, could you please check whether it works?

https://github.com/clamor-s/morpheus_kernel_asus_tf201/blob/asus/arch/arm/mach-tegra/board-cardhu-panel.c#L315

```patch
-		regulator-enable-ramp-delay = <300000>;
+		regulator-enable-ramp-delay = <20000>;
```

maximschwalm:
> The original downstream kernel seems use 20ms delay, could you please check whether it works?
> 
> https://github.com/clamor-s/morpheus_kernel_asus_tf201/blob/asus/arch/arm/mach-tegra/board-cardhu-panel.c#L315
> 
> ```diff
> -		regulator-enable-ramp-delay = <300000>;
> +		regulator-enable-ramp-delay = <20000>;
> ```

Shouldn't startup-delay-us be used when the regulator is fixed?

digetx:
Looks like both should work equally. I don't know why `startup-delay-us` exists, maybe it's a legacy property.

maximschwalm:
No, it doesn't work. Same issue again

digetx:
Can you try a lower delay? Like 1ms.

arch/arm/boot/dts/tegra30-asus-transformer-common.dtsi
clamor-s: Michal dts may not cover all t30 transformers. Currently regulators are set according to downstream.

clamor-s: Michal dts may not cover all t30 transformers. Currently regulators are set according to downstream.

clamor-s:
@digetx may you halt this PR till 9th of June. I would really like to review them as well. I'm currently out of home network with poot wifi.

digetx:
We can wait. I want to know why regulator delay makes a difference, this is the biggest question for me here.

arch/arm/boot/dts/tegra30-asus-transformer-common.dtsi
digetx: It's likely that those regulators indeed should be the same for all, but it's better to ensure.

maximschwalm:
After further testing it seems like removing the ramp delay didn't fix it at all. It was probably just a coincidence that the display worked after reboot, but I could reboot several times in row and it was still working.

okias:
> ...but I could reboot several times in row and it was still working.

maybe you set something before and it's not (properly) initialized by kernel?

digetx:
@maximschwalm could you please check reboot with this change?

```patch
diff --git a/drivers/gpu/drm/panel/panel-simple.c b/drivers/gpu/drm/panel/panel-simple.c
index a46f86d2fea5..5bbf5132dc41 100644
--- a/drivers/gpu/drm/panel/panel-simple.c
+++ b/drivers/gpu/drm/panel/panel-simple.c
@@ -4954,7 +4954,7 @@ static struct mipi_dsi_driver panel_simple_dsi_driver = {
 	},
 	.probe = panel_simple_dsi_probe,
 	.remove = panel_simple_dsi_remove,
-	.shutdown = panel_simple_dsi_shutdown,
+// 	.shutdown = panel_simple_dsi_shutdown,
 };
 
 static int __init panel_simple_init(void)

```

maximschwalm:
> @maximschwalm could you please check reboot with this change?
Thanks, this seems to fix it

digetx:
It may lead to the root of the problem, but it's not a fix yet. Could be that the W1 gpio needs to be enabled for the TC bridge, have you checked it?

```diff
diff --git a/arch/arm/boot/dts/tegra30-asus-tf700t.dts b/arch/arm/boot/dts/tegra30-asus-tf700t.dts
index 1dec83d07555..bec5056efb66 100644
--- a/arch/arm/boot/dts/tegra30-asus-tf700t.dts
+++ b/arch/arm/boot/dts/tegra30-asus-tf700t.dts
@@ -157,7 +157,6 @@ display-panel@1 {
 				compatible = "panasonic,vvx10f004b00";
 				reg = <1>;
 
-				power-supply = <&vdd_pnl>;
 				backlight = <&backlight>;
 
 				port {
@@ -231,7 +230,7 @@ vdd_1v2_mipi: regulator@6 {
 		regulator-boot-on;
 		gpio = <&gpio TEGRA_GPIO(BB, 3) GPIO_ACTIVE_HIGH>;
 		enable-active-high;
-		vin-supply = <&vdd_3v3_sys>;
+		vin-supply = <&vdd_pnl>;
 	};
 
 	vibrator {

```

We shouldn't remove the W1 ramp delay (maybe change it to 20msec) and should be better not to change the limits of those regulators where we not sure yet. Could you please update this PR?

maximschwalm:
> It may lead to the root of the problem, but it's not a fix yet. Could be that the W1 gpio needs to be enabled for the TC bridge, have you checked it?

Not yet. Should I try it without the previous patch?

digetx:
Yes, without the previous patch.

digetx:
The reboot problem is unrelated to the rest of the changes in this PR, it shouldn't block fixing the MMC regulators.

maximschwalm:
> Yes, without the previous patch.

This time it fails sooner and doesn't gets fixed by another reboot

digetx:
Now you can narrow down the panel_simple_dsi_shutdown() to the point that breaks display after reboot. Once we will know that point, we will think about how to fix it.

The updated PR looks good to me. I'll merge it, thanks!

PR #49, opened 2021-07-07
WIP: ARM: tegra: Surface RT devicetree update

CrackTheSurface:
# uSD Card:
added CardDetection and WriteProtect gpio. They are the same pins as for Cardhu

# USB Vbus regulator
pulling the gpio EE0 Low disconnects all USB devices.

# Lid halleffect sensor
closing the lid triggers a hall effect sensor which will suspend the device after 3seconds.

# NCT1008 temperature sensor
Same sensor is used in Cardhu and Ouya.
lm-sensor read the correct temperatures

# tCover moved to efi.dts
tCover depends on a PMIC regulator and an I2C power up sequence provided by UEFI Firmware.
For APX tCover support further steps are necessary.


digetx:
Thanks!

PR #50, opened 2021-07-09
CI: github: deploy

okias:
This should be making tuxmake builds for each push on master branch.

Each time is generated tagged release which overwrites previous.
This tagged release contains tuxmake archive (zImage, dtbs, vmlinux etc.)

Based on: https://gitlab.com/okias/kernel-simpleci


digetx (approved):
Looks good, thank you!

PR #51, opened 2021-07-12
Transformers update

clamor-s:


arch/arm/boot/dts/tegra30-asus-transformer-common.dtsi
```diff
@@ -1390,7 +1390,6 @@
 		regulator-boot-on;
 		gpio = <&pmic 6 GPIO_ACTIVE_HIGH>;
 		enable-active-high;
-		vin-supply = <&vdd_5v0_bat>;
```
digetx, line 1393: Is anything wrong here?

digetx:
@davidhildenbrand Looks like it tries to make a release tag here.

arch/arm/boot/dts/tegra30-asus-transformer-common.dtsi
clamor-s: 3.3v according to schematis is not supplied by anything and starts same as 5v bat and they are not connected

digetx: The TF700T pdf says that the 3VSUS is sourced from the battery supply. I assume this should be the same for all transformers.

clamor-s: [https://github.com/clamor-s/files/blob/main/ASUS%20Pegatron%20TF201X%20TF300T_RG.pdf](second page)

clamor-s: yes, but 5v bat is not battery supply. 5v sys and 5v sby descends from it, but not 3.3v

digetx: The schematics clearly shows that 3.3 and 5 are generated from AC_BAT_SYS by the TPS charger IC.

clamor-s: ac-bat-sys = vdd-ac-bat?

digetx: yes

clamor-s: Ok, will update.

arch/arm/boot/dts/tegra30-asus-transformer-common.dtsi
```diff
@@ -1230,9 +1230,9 @@
 				};
 
 				vdd2_reg: vdd2 {
-					regulator-name = "vddio_ddr";
-					regulator-min-microvolt = <600000>;
-					regulator-max-microvolt = <1500000>;
+					regulator-name = "vdd_1v2_gen";
+					regulator-min-microvolt = <1200000>;
+					regulator-max-microvolt = <1200000>;
```
digetx, line 1235: Could you please tell where from you got these names and values?

clamor-s: https://github.com/grate-driver/linux/blob/master/arch/arm/boot/dts/tegra30-asus-nexus7-grouper-ti-pmic.dtsi#L44

digetx: Have you checked it on all transformer devices?

clamor-s: Yes, tf201/tf300t/tf700t both native and kexec. Furthermore, thos values can be logged from downstream is /sys/

maximschwalm: Weren't the names correct?

clamor-s: @digetx is value 8 legit for ti,regulator-ext-sleep-control

clamor-s:
@digetx looks like tf701t dock actually is a i2c device and CAN be described as asus,dock-ec. I;d ask you to revert part of changes I've proposed. Me and jenneron are currently working on proper solution that does not need specific per-device data definitions.

drivers/mfd/asus-ec.c
```diff
@@ -515,9 +480,12 @@ static int asus_ec_remove(struct i2c_client *client)
 	struct asus_ec_data *priv = i2c_get_clientdata(client);
 
 	mfd_remove_devices(&priv->self->dev);
+
 	sysfs_remove_link(&client->dev.kobj, "dockram");
 	sysfs_remove_group(&client->dev.kobj, &asus_ec_attr_group);
 
+	priv->logging_disabled = false;
```
digetx, line 487: The driver is removed, logging_disabled should already be false and it won't do anything useful anyways. This line is unneeded.

clamor-s: Ok, I doubted on this. Thanks, will remove.

digetx (commented):
TF101 would never boot previously. if it required firmware. Downstream kernels may have TF code, but it should be disabled by default. T20 consumer devices don't use TF in general.

arch/arm/boot/dts/tegra30-asus-transformer-common.dtsi
digetx: Likely that the names were correct and N7 has incorrect names, but I'm not 100% sure.

clamor-s:
> TF101 would never boot previously. if it required firmware. Downstream kernels may have TF code, but it should be disabled by default. T20 consumer devices don't use TF in general.

Agree, there are some thoughts about booting with android bootloader. TF101G does not have nvflash ability.

May this work? https://github.com/torvalds/linux/blob/master/arch/arm/boot/dts/tegra114-tn7.dts#L14-L16
Jenneron proposed to check, but none of TF101 users provided addresses needed.

digetx:
Missing initrd shouldn't prevent kernel from booting. The problem that we had on N7 is that bootloader overwrites part of kernel with initrd if kernel image is too big. But it could be more complicated, like an odd h/w configuration left after bootloader. It's possible to debug kernel booting blindly by inserting PMC reboot call for finding the place in the code that causes the hang.

Enabling firmware for TF101 should break it entirely, please don't do it. I suppose that the TF patch wasn't tested, correct?

clamor-s:
Actually it was and it worked. At least it did not brake anything. I'll call tf101 users in pmos on transformer channel for more info.

drivers/power/supply/asus-ec-battery.c
```diff
@@ -237,14 +236,14 @@ static int asusec_battery_probe(struct platform_device *pdev)
...
-	priv->ec_addr = pdata->ec_addr;
+	priv->ec_addr = 0x14;
 	priv->batt_data_ts = jiffies - 1;
 	priv->last_state = POWER_SUPPLY_STATUS_UNKNOWN;
 
 	cfg.of_node = pdev->dev.parent->of_node;
 	cfg.drv_data = priv;
 
-	if (pdata->is_pad)
+	if (!of_property_match_string(cfg.of_node, "compatible", "asus,pad-ec"))
```
digetx, line 246: Couldn't `of_device_is_compatible()` be use here?

digetx:
Picasso doesn't have such problem.

drivers/power/supply/asus-ec-battery.c
digetx: I'll change it after merging. The rest looks okay.

digetx:
Thanks!

PR #53, opened 2021-07-13
ARM: configs: Rebuild tegra_defconfig

maximschwalm:
- Rebuild tegra_defconfig
- Make the switch to the new CROS_EC options


digetx:
Should be a better to fix the CONFIG_CROS_EC at first and then rebuild the config in a separate patch. The EC patch should contain the fixes tag pointing at 67c70aa86f8b796c45a08e0c955dabd826f6d0c6.

arch/arm/configs/tegra_defconfig
```diff
@@ -118,8 +118,8 @@ CONFIG_INPUT_JOYDEV=y
 CONFIG_INPUT_EVDEV=y
 CONFIG_KEYBOARD_GPIO=y
 CONFIG_KEYBOARD_TEGRA=y
-CONFIG_KEYBOARD_CAP11XX=y
 CONFIG_KEYBOARD_CROS_EC=y
+CONFIG_KEYBOARD_CAP11XX=y
```
digetx, line 116: Shouldn't this belong to the rebuilding patch?

maximschwalm: OK, I put this change in the next commit.

digetx:
The `fixes` tag shouldn't be wrapped, I'll correct it after merging. Thanks!

PR #54, opened 2021-07-13
Update Chagall

clamor-s:
I made a battery driver. It is not finished but is in stable state, functions as intended. Additionally updated chagall dts.

### DIFF
--- a/arch/arm/boot/dts/tegra30-pegatron-chagall.dts
+++ b/arch/arm/boot/dts/tegra30-pegatron-chagall.dts
@@ -1098,7 +1098,13 @@
 		status = "okay";
 		clock-frequency = <400000>;
 
-		/* cg7153am battery driver is here */
+		fuel-gauge@10 {
+			compatible = "cg7153am,battery";
+			reg = <0x10>;
+
+			monitored-battery = <&battery>;
+			power-supplies = <&mains>;
+		};
 
 		/* Wolfson Microelectronics WM8903 audio codec */
 		wm8903: audio-codec@1a {
@@ -1493,6 +1499,19 @@
 
 		vmmc-supply = <&vdd_usd>;	/* ldo2 */
 		vqmmc-supply = <&vddio_usd>;	/* ldo3 */
+	};
+
+	battery: battery {
+		compatible = "simple-battery";
+		charge-full-design-microamp-hours = <3050000>;
+		energy-full-design-microwatt-hours = <23000000>;
+		operating-range-celsius = <0 45>;
+	};
+
+	mains: ac-adapter-detect {
+		compatible = "gpio-charger";
+		charger-type = "mains";
+		gpios = <&gpio TEGRA_GPIO(V, 1) GPIO_ACTIVE_HIGH>;
 	};
 
 	brcm_wifi_pwrseq: wifi-pwrseq {
--- a/arch/arm/boot/dts/tegra30-pegatron-chagall.dts
+++ b/arch/arm/boot/dts/tegra30-pegatron-chagall.dts
@@ -89,7 +89,7 @@
 			status = "okay";
 
 			hdmi-supply = <&hdmi_5v0_sys>;
-			pll-supply = <&vdd_1v8>;
+			pll-supply = <&vdd_1v8_vio>;
 			vdd-supply = <&vdd_3v3_sys>;
 
 			nvidia,hpd-gpio = <&gpio TEGRA_GPIO(N, 7) GPIO_ACTIVE_HIGH>;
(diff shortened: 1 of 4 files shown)


arch/arm/boot/dts/tegra30-pegatron-chagall.dts
```diff
@@ -1540,30 +1558,23 @@
...
-		vqmmc-supply = <&vdd_1v8>;
+		vqmmc-supply = <&vdd_1v8_vio>;
 		non-removable;
 	};
 
 	usb@7d000000 {
 		compatible = "nvidia,tegra30-udc";
 		status = "okay";
-		dr_mode = "peripheral";
+		dr_mode = "otg";
```
digetx, line 1568: OTG mode should require presence of the VBUS regulator. Please don't enable it since it should break USB port if not done properly. Was this PR fully tested?

clamor-s: Yes, otg work. Vbus is set in usb-phy@7d000000

digetx: VBUS must be set in the USB controller node because CI driver is responsible for the mode switching. Setting VBUS in the PHY node should break USB peripheral mode. A working USB should mean that the VBUS regulator is wrong. Peviously it was incorrect too because VBUS isn't needed for the peripheral mode.

The VBUS regulator must be OFF for the peripheral mode and ON for the host mode.

I'm doubting that microUSB supports OTG here at all since OTG usually not wired in hardware if there is a proper host port.

clamor-s: vbus is not needed for peripheral, but warnings of missing regulator popped

digetx: Message about a dummy regulator isn't a error.

Since you specified the VBUS regulator in the PHY node, it means that VBUS is enabled by the PHY driver permanently. The peripheral mode can't work with the enabled VBUS. If peripheral mode works, it means that the DD3 gpio VBUS regulator is wrong.

arch/arm/boot/dts/tegra30-pegatron-chagall.dts
```diff
@@ -1248,27 +1258,36 @@
...
 				};
 
-				vdd2_reg: vdd2 {
-					regulator-name = "vdd2_1v2";
-					regulator-min-microvolt = <600000>;
-					regulator-max-microvolt = <1500000>;
+				vddio_1v2_ddr: vdd2 {
+					regulator-name = "vddio_1v2_ddr";
+					regulator-min-microvolt = <1200000>;
+					regulator-max-microvolt = <1200000>;
```
digetx, line 1290: Have you checked that it ineed should be 1.2v here?

clamor-s: Yes, according tf300t schematics (chagall derives from it) and downstream logs.

digetx: Alright

arch/arm/boot/dts/tegra30-pegatron-chagall.dts
clamor-s: You are correct. Chagall `tegra_ehci_utmi_pdata ` match transformers but have vbus_reg.

digetx: The default mode for microUSB port is the peripheral mode. If microUSB has ID pin connected to SoC and it has VBUS regulator, then it may support OTG mode. Transformers should use peripheral mode.

Looking at the description of the Chagall, I see that it has OTG-capable microUSB + GSM modem. Hence it should be the peripheral mode for the microUSB until the OTG will be properly implemented and fully tested.

clamor-s: transformers in usb-phy use peripheral (remove last commit, correct)? Chagall as well till fixed?

digetx: Yes, you should use peripheral mode for UDC @7d000000 for now. The dummy vbus-supply should be removed from the UDC PHY for completeness.

So here, for Chagall, you should revert the 7d000000 changes and remove the vbus-supply. If @cunidev has OTG adapter, then we may try to enable and test OTG later on.

clamor-s: Transformer changes dropped. May vbus regulators stay so warnings about dummy regulators are not produced? Transformers have no regularoers/gpios that control otg and classic otg itself unlike other devices.

Chagall changes reverted but regulators I left. If you find those impropper, remove them on squasing commits. Thank you!

PR #55, opened 2021-07-13
CI: implement checkpatch.pl

okias:


digetx:
Which cmdline arguments are used for running the checkpatch.pl? Do we use `--strict` option?

okias:
I didn't found, that checkpatch is run with --strict option, do you want it to be run with --strict?

digetx:
Yes, please add --strict.

okias:
@digetx done

digetx:
Very nice, thank you!

PR #56, opened 2021-07-13
FIXUP: speed-up CI with SCCache (thx Mozilla!)

okias:
Also drop ccache from dependencies and separate them into own block


digetx:
Thanks!

PR #58, opened 2021-07-14
WIP: ARM: tegra: surface-rt-efi: disable LP2

jenneron:
Now suspend partially (of course CPU voltage is enabled) works, although this is not enough to get working CPU idle. Device used to hang while suspending before this change.

BTW thanks for the tip.


jenneron:
I also configured windows button to wake up the device as it is configured this way in stock windows

digetx:
Thanks!

PR #59, opened 2021-07-19
ARM: tegra: Surface RT: battery support

CrackTheSurface:
Custom driver derived from original APCI SSDT table.
Works for APX and UEFI.
Use CONFIG_BATTERY_SURFACE_RT=y.

Surface RT uses an EmbeddedController which is accessed as ACPI control method battery.
The driver imitates the _BST and _BIX methods.

Signed-off-by: Jonas Schwöbel <jonasschwoebel@yahoo.de>

### DIFF
--- a/drivers/rtc/rtc-tps65910.c
+++ b/drivers/rtc/rtc-tps65910.c
@@ -467,6 +467,6 @@ static struct platform_driver tps65910_rtc_driver = {
 };
 
 module_platform_driver(tps65910_rtc_driver);
-MODULE_ALIAS("platform:rtc-tps65910");
+MODULE_ALIAS("platform:tps65910-rtc");
 MODULE_AUTHOR("Venu Byravarasu <vbyravarasu@nvidia.com>");
 MODULE_LICENSE("GPL");


MAINTAINERS
```diff
@@ -12374,6 +12374,13 @@ L:	platform-driver-x86@vger.kernel.org
 S:	Supported
 F:	drivers/platform/surface/surfacepro3_button.c
 
+MICROSOFT SURFACE RT
+M:	Jonas Schwöbel <jonasschwoebel@yahoo.de>
+S:	Supported
+F:	drivers/power/supply/surface-rt-battery.c
+F:	arch/arm/boot/dts/tegra30-microsoft-surface-rt.dts
+F:	arch/arm/boot/dts/tegra30-microsoft-surface-rt-efi.dts
```
jenneron, line 12382: I don't think we will upstream -efi device-tree

drivers/power/supply/surface-rt-battery.c
```diff
@@ -0,0 +1,196 @@
...
+											enum power_supply_property psp,
+											union power_supply_propval *val)
+{
+	struct srt_battery_device *srt_battery = power_supply_get_drvdata(psy);
+	struct i2c_client *client = srt_battery->client;
+
+	static char strBuf[8];
+
+	gpiod_set_value(srt_battery->powerdown_gpio, 0); // power up
+	usleep_range(1000, 1500); // 1ms from ACPI wait till IC is ready
```
digetx, line 63: Why is it a `powerdown` gpio and not `powerup`? Do you know for sure that the enable state is physically inverted in h/w or what is the reason behind it?

Why gpio can't be enabled once at a driver probe time? This function is invoked for each property whenever supply state changes, so it looks sub-optimal to re-enable it each time.

drivers/power/supply/surface-rt-battery.c
```diff
@@ -0,0 +1,196 @@
...
+	int ret = -1;
+
+	srt_battery = devm_kzalloc(dev, sizeof(*srt_battery), GFP_KERNEL);
+	if (!srt_battery)
+		return -ENOMEM;
+
+	srt_battery->client = client;
+	srt_battery->dev = dev;
+
+	srt_battery->powerdown_gpio = devm_gpiod_get_optional(srt_battery->dev, "powerdown", GPIOD_OUT_LOW);
```
digetx, line 149: Why gpio is optional?

drivers/power/supply/surface-rt-battery.c
CrackTheSurface: > Why is it a powerdown gpio and not powerup?

I looked through some devicetrees and only found powerdown and enable. powerdown fitted better imo (read further)...

Gpio set to High: i can read from the device
Gpio set to Low: device doesn't show up on i2cdetect. (Is this a problem for the driver? for probing? It worked for me but that doesn't mean that it's the right way)
ACPI also describes this gpio as PowerResource. It might be a regulator.
But there is no datasheet. All information is taken from ACPI table.


> Why gpio can't be enabled once at a driver probe time?

It could, but
ACPI does the same as my driver. It sets the GPIO to high. Waits 1ms. Reads all the values at once. Sets the GPIO to low.
I guess it has something to do with powersaving. That's why i choose powerdown

drivers/power/supply/surface-rt-battery.c
CrackTheSurface: it shouldn't be optional.
First time i am doing stuff like this. Didn't thought about testing it without GPIO. Would have clearly shown that this doesn't work for APX where the Pin needs to be set to HIGH. UEFI already does this.
I will replace it with `devm_gpiod_get(...)`

drivers/power/supply/surface-rt-battery.c
digetx: So high=enabled, which is natural for the active-high enable-gpio. I don't see why powerdown is better, it's logically inverted.

I2C devices are discovered from the device-tree description, gpio doesn't need to be enabled for that.

digetx:
Does battery support interrupt notification about changed capacity? Will you add it later on? Or will you add software polling?

CrackTheSurface:
> Does battery support interrupt notification about changed capacity? Will you add it later on? Or will you add software polling?

There is no hardware mechanism i am aware of.
From ACPI spec:
>"When the present state of the battery has changed or when the trip point set by the _BTP control 
method is reached or crossed, the hardware will assert a general purpose event. The AML code 
handler for this event issues a `Notify(battery_device, 0x80)` on the battery device"

We dont have a `Notify(battery_device, 0x80)` so i guess we need to poll for changes.

How do i notify the system that the capacity has reached a certain level?

digetx:
See drivers that use schedule_delayed_work(), like acer_a500_battery.c. You need to notify when capacity value changed, not when it reached a certain level.

digetx:
@CrackTheSurface Will you update this PR? I can merge it as-is and then make the necessary changes locally.

CrackTheSurface:
I can/will update it but I am pretty busy right. If it can wait until next week i will do it.

digetx:
It can wait, thanks.

PR #61, opened 2021-07-25
WIP: chagall-dts: add regulators for 3G modem

cunidev:
This device is based on a Huawei MU739 USB modem, which requires several GPIOs and extra regulators for initialization.

Signed-off-by: Raffaele Tranquillini <raffaele.tranquillini@gmail.com>

### DIFF
--- a/arch/arm/boot/dts/tegra30-pegatron-chagall.dts
+++ b/arch/arm/boot/dts/tegra30-pegatron-chagall.dts
@@ -97,7 +97,19 @@
 		};
 	};
 
-	gpio@6000d000 {};	/* set up from downstream if necessary */
+	gpio@6000d000 {
+		init-mode-3g-hog {
+			gpio-hog;
+			gpios =	<TEGRA_GPIO(P, 0) GPIO_ACTIVE_HIGH>, // CHAGALL_3G_RESIN_N
+				<TEGRA_GPIO(V, 7) GPIO_ACTIVE_HIGH>, // CHAGALL_3G_PWRDWN_N
+				<TEGRA_GPIO(V, 6) GPIO_ACTIVE_HIGH>, // CHAGALL_3G_DISABLE_N
+				<TEGRA_GPIO(O, 2) GPIO_ACTIVE_HIGH>, // CHAGALL_3G_R_ON
+				<TEGRA_GPIO(X, 5) GPIO_ACTIVE_HIGH>, // CHAGALL_3G_RESOUT
+				<TEGRA_GPIO(N, 2) GPIO_ACTIVE_HIGH>, // CHAGALL_3G_WW_WAKE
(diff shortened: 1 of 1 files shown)


cunidev:
This is a work-in-progress, not ready or full yet. It is aimed at making gemini closer to tilapia, and to 3G modem support once a driver enabling it is written.

arch/arm/boot/dts/tegra30-pegatron-chagall.dts
```diff
@@ -1577,8 +1598,17 @@
...
+	};
+
+	usb-phy@7d004000 {
+		status = "okay";
+		vbus-supply = <&vdd_vbus_3g>;
+	};
+	
 	usb@7d008000 {
-		status = "okay";		
+		status = "okay";
```
digetx, line 1611: `status` should be disabled until it works

arch/arm/boot/dts/tegra30-pegatron-chagall.dts
```diff
@@ -1462,6 +1474,15 @@
...
 	};
+	
+	vdd_vbus_3g: regulator@6 {
+		compatible = "regulator-fixed";
+		regulator-name = "vdd_vbus_3g";
+		regulator-min-microvolt = <3800000>;
+		regulator-max-microvolt = <3800000>;
+		gpio = <&gpio TEGRA_GPIO(K, 7) GPIO_ACTIVE_HIGH>;
+		enable-active-high;
+	};
```
digetx, line 1485: It looks more like a 3g modem power than the USB VBUS, where did you got the VBUS?

digetx: Could you point me at the downstream chagall kernel that has a working 3g?

cunidev: The "hog" is defined here: https://github.com/n5047c/android_kernel_chagall/blob/master/arch/arm/mach-tegra/board-chagall-baseband.c#L44

And the regulator for enabling modem (so not vbus) here: https://github.com/n5047c/android_kernel_chagall/blob/master/arch/arm/mach-tegra/board-chagall-power.c#L590

digetx: Looks like it's UTMI and not HSIC USB modem, so it may actually work if there is a driver for that modem.

cunidev: How should the 3v8 regulator be declared in this case?

digetx: Not sure for now.  We can take a closer look at how to implement it properly once you will have a working draft, i.e. something that can make a call or connect to internet.

PR #62, opened 2021-07-26
Update Transformers and Chagall

clamor-s:
TRANSFORMERS
- vdd1 regulator is removed as it is not used on transformers
- added 5v_sys reg in usb1 accorging to schematics 
- added max current prop in tf battery driver
- re-grouped tf pinmux (cosmetics, no value changes) and commented wm8903 gpio-cfg
- max battery temp multiplied by 10 since it gave me in log 4.5C not 45C
- blocked power supply event from USB (caused device wakeups when plugged on Max TF700T) - may you add commit msg into battery driver commit msg since it may be important

CHAGALL
- vdd1 regulator is removed as it is not used on chagall
- battery driver was updated and now has power led indication support like in downstream
- added extcon gpio keys group (dock insert detection and lineout detection)
- chagall may be removed from WIP since it is mostly complete now

### DIFF
--- a/arch/arm/boot/dts/tegra30-asus-tf300t.dts
+++ b/arch/arm/boot/dts/tegra30-asus-tf300t.dts
@@ -120,11 +120,11 @@
 			micdet-delay = <100>;
 
 			gpio-cfg = <
-				0xffffffff
-				0xffffffff
-				0x00000000 
-				0xffffffff /* Interrupt, output */
-				0xffffffff /* BCLK, input, active high */
+				0xffffffff /* don't touch */
+				0xffffffff /* don't touch */
+				0x00000000 /* Speaker-enable GPIO, output, low */
+				0xffffffff /* don't touch */
+				0xffffffff /* don't touch */
 			>;
 
 			AVDD-supply  = <&vdd_1v8_vio>;
--- a/arch/arm/boot/dts/tegra30-asus-tf300t.dts
+++ b/arch/arm/boot/dts/tegra30-asus-tf300t.dts
@@ -134,6 +134,11 @@
 		};
 	};
 
+	/* TF300T does not support SDR104 mode */
+	mmc@78000000 {
+		max-frequency = <104000000>;
+	};
+
 	pad_battery: pad-battery {
 		compatible = "simple-battery";
 		charge-full-design-microamp-hours = <2940000>;
--- a/drivers/power/supply/asus-ec-battery.c
+++ b/drivers/power/supply/asus-ec-battery.c
@@ -21,6 +21,8 @@
 #define ASUSEC_BATTERY_FULL_CHARGED		0x20
 #define ASUSEC_BATTERY_FULL_DISCHARGED		0x10
 
+#define ASUSEC_CHARGER_USB_MASK			0x43
+
 #define TEMP_CELSIUS_OFFSET			2731
 
 struct asusec_battery_data {
@@ -180,6 +182,23 @@ static int pad_battery_get_property(struct power_supply *psy,
 	return 0;
 }
 
+static int asusec_battery_no_usb(struct asusec_battery_data *priv)
+{
+	int ret = 0;
+	u8 buf[DOCKRAM_ENTRY_BUFSIZE];
+	char *plug = buf;
+
+	ret = asus_dockram_read(priv->ec->dockram, 0x0A, plug);
+	if (ret < 0)
+		return -EINVAL;
+
+	ret = plug[1] & ASUSEC_CHARGER_USB_MASK;
+	if (ret == ASUSEC_CHARGER_USB_MASK)
+		return 0;
+
+	return ret;
+}
+
 static void asusec_battery_poll_work(struct work_struct *work)
 {
 	struct asusec_battery_data *priv =
@@ -197,7 +216,7 @@ static void asusec_battery_poll_work(struct work_struct *work)
 	else
 		state = POWER_SUPPLY_STATUS_CHARGING;
 
-	if (priv->last_state != state) {
+	if (priv->last_state != state && asusec_battery_no_usb(priv)) {
 		priv->last_state = state;
 		power_supply_changed(priv->battery);
 	}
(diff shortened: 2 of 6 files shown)


arch/arm/boot/dts/tegra30-asus-tf300t.dts
```diff
@@ -138,14 +138,14 @@
...
-		operating-range-celsius = <0 45>;
+		operating-range-celsius = <0 450>;
 	};
 
 	dock_battery: dock-battery {
 		compatible = "simple-battery";
 		charge-full-design-microamp-hours = <2260000>;
 		energy-full-design-microwatt-hours = <16000000>;
-		operating-range-celsius = <0 45>;
+		operating-range-celsius = <0 450>;
```
digetx, line 148: 450?

clamor-s: 45°С else I get 4.5°C in log

digetx: Shouldn't it be multiplied in the driver?

clamor-s: Hm, I should add it in asus-ec-battery?

digetx: Yes

clamor-s: Ok, will do for both transformers and chagall

digetx:
Is it ready? The temp is unchanged now, but I can fix it after merge.

clamor-s:
You may check your devices with lm_sensors

digetx:
It works, what should I check?

clamor-s:
`pad_battery-isa-0000
Adapter: ISA adapter
in0:           7.63 V  (max =  +8.40 V)
temp:         +23.5°C  (low  =  +0.0°C, high =  **+4.5°C**)
curr1:         1.13 A  (max =  +1.64 A)`

drivers/mfd/asus-ec.c
```diff
@@ -393,7 +397,8 @@ static ssize_t ec_irq_store(struct device *dev,
 	struct asusec_info *ec = dev_get_drvdata(dev);
 	struct asus_ec_data *priv = to_ec_data(ec);
 
-	irq_wake_thread(priv->self->irq, priv);
+	if (priv->self->irq)
+		irq_wake_thread(priv->self->irq, priv);
```
digetx, line 401: This NULL check should be unnecessary.

drivers/mfd/asus-ec.c
```diff
@@ -130,6 +130,10 @@ static int asusec_signal_request(const struct asusec_info *ec)
 {
 	struct asus_ec_data *priv = to_ec_data(ec);
 
+	/* if request gpio is not set, skip signal request */
+	if (!priv->ecreq)
+		return 0;
+
```
digetx, line 136: This NULL check is unnecessary.

drivers/mfd/asus-ec.c
clamor-s: Will irq_wake_thread handle it?

digetx: Yes

drivers/mfd/asus-ec.c
clamor-s: gpiod_set_value_cansleep will give dev_warnings if NULL

digetx: It won't.

clamor-s: I'm still thinking if those irq and gpio commits are needed. TF701 dock is a tuff piece and does not works.

digetx: It should be better to add comment to `devm_gpiod_get_optional`, explaining why it's optional.

clamor-s: Ok, I'll remove those checks

clamor-s:
@digetx https://github.com/grate-driver/linux/blob/master/Documentation/devicetree/bindings/power/supply/maxim,max17042.yaml#L38

digetx:
That is not a standard property. You can change the default properties to the tenths since nobody really uses them, but then it needs to be done properly by updating the binding and device-trees, you should also send those patches to upstream.

clamor-s:
Actually all that is already done since everywhere it is set as degree not a tenth of degree

clamor-s:
ok, I will update battery driver but with note about degrees

clamor-s:
We are getting some strange stuff from nct1008 thermal zones.

arch/arm/boot/dts/tegra30-asus-tf300t.dts
```diff
@@ -134,6 +134,11 @@
 		};
 	};
 
+	/* TF300T does not support SDR104 mode */
+	mmc@78000000 {
+		max-frequency = <104000000>;
+	};
```
digetx, line 140: If it doesn't support 104MHz, then why the limit is set to 104MHz?

drivers/power/supply/asus-ec-battery.c
```diff
@@ -129,12 +133,13 @@ static int pad_battery_get_property(struct power_supply *psy,
...
 
+	/* Temperature from dts is read in decidegree C */
 	case POWER_SUPPLY_PROP_TEMP_MIN:
-		val->intval = battery_info.temp_min;
+		val->intval = battery_info.temp_min * 10;
 		break;
 
 	case POWER_SUPPLY_PROP_TEMP_MAX:
-		val->intval = battery_info.temp_max;
+		val->intval = battery_info.temp_max * 10;
```
digetx, line 142: The min/max checks are missing, but I can fix it after applying,

digetx:
tf700t is 30C, I don't see anything overheating.

arch/arm/boot/dts/tegra30-asus-tf300t.dts
clamor-s: SDR104 is 208MHz

drivers/power/supply/asus-ec-battery.c
clamor-s: Ok, would be nice. Chagall battery has similar setup.

clamor-s:
I forgot Andy

digetx:
I don't know what's wrong with Andy's setup, that's not enough information.

The PR look okay, thanks!

PR #63, opened 2021-07-28
WIP: ARM: tegra: surface-rt: add WiFi/BT

jenneron:
WiFi asks for correct firmware and works with `linux-firmware-mrvl` installed.

Actual chip is 8797. Bluetooth boots with errors about missing firmware. It wants `helper_uart_3000000.bin` and when using the helper which i found for another chip it asks for `mrvl/uart8897_bt.bin` which is wrong (because of wrong helper?) and which i have not found as well as 8797. Also according to ACPI it has interrupt and some additional GPIO, but there are no those in Linux driver.


jenneron:
I think we can merge this. Although there are some things to do: test in APX + uboot variant, find tps regulators for sdio, find (ask marvell?) correct firmware for BT. We can do these things later.

digetx:
Thanks!

PR #65, opened 2021-08-09
ARM: Tegra: Surface RT: Battery and Charger support

CrackTheSurface:
Battery and Charger support for UEFI and APX boot.
use CONFIG_BATTERY_CHARGER_SURFACE_RT to enable support.

Signed-off-by: Jonas Schwöbel <jonasschwoebel@yahoo.de>


digetx:
Thanks!

PR #67, opened 2021-08-13
Asus Transformers Updates

clamor-s:
- add TF300TG
- switch to simple panels
- improve asus-ec
- add schema yamls

### DIFF
--- a/arch/arm/boot/dts/tegra30-asus-tf201.dts
+++ b/arch/arm/boot/dts/tegra30-asus-tf201.dts
@@ -5,6 +5,7 @@
 /* CPU Speedo ID 3, Soc Speedo ID 2, CPU Process: 1, Core Process: 0 */
 
 #include "tegra30-asus-transformer-common.dtsi"
+#include "tegra30-asus-lvds-display.dtsi"
 
 / {
 	model = "Asus Transformer Prime TF201";
@@ -123,65 +124,8 @@
 		operating-range-celsius = <0 45>;
 	};
 
-	/* HannStar HSD101PWW2 Rev0-A00/A01 LCD SuperIPS+ HD panel */
 	display-panel {
-		compatible = "panel-lvds";
-
-		power-supply = <&vdd_pnl>;
-		backlight = <&backlight>;
-
-		width-mm = <217>;
-		height-mm = <136>;
-
-		data-mapping = "jeida-24";
-
-		panel-timing {
-			/* 1280x800@60Hz */
-			clock-frequency = <68000000>;
-			hactive = <1280>;
-			vactive = <800>;
-			hfront-porch = <48>;
-			hback-porch = <18>;
-			hsync-len = <30>;
-			vsync-len = <5>;
-			vfront-porch = <3>;
-			vback-porch = <12>;
-		};
-
-		port {
-			panel_input: endpoint {
-				remote-endpoint = <&bridge_output>;
-			};
-		};
-	};
-
-	/* Texas Instruments SN75LVDS83B LVDS Transmitter */
-	lvds-encoder {
-		compatible = "ti,sn75lvds83", "lvds-encoder";
-
-		powerdown-gpios = <&gpio TEGRA_GPIO(N, 6) GPIO_ACTIVE_LOW>;
-		power-supply = <&vdd_3v3_sys>;
-
-		ports {
-			#address-cells = <1>;
-			#size-cells = <0>;
-
-			port@0 {
-				reg = <0>;
-
-				bridge_input: endpoint {
-					remote-endpoint = <&dpi_output>;
-				};
-			};
-
-			port@1 {
-				reg = <1>;
-
-				bridge_output: endpoint {
-					remote-endpoint = <&panel_input>;
-				};
-			};
-		};
+		compatible = "hannstar,hsd101pww2";
 	};
 
 	vibrator {
--- a/arch/arm/boot/dts/tegra30-pegatron-chagall.dts
+++ b/arch/arm/boot/dts/tegra30-pegatron-chagall.dts
@@ -11,6 +11,7 @@
 #include "tegra30.dtsi"
 #include "tegra30-cpu-opp.dtsi"
 #include "tegra30-cpu-opp-microvolt.dtsi"
+#include "tegra30-asus-lvds-display.dtsi"
 
 / {
 	model = "Pegatron Chagall based tablet";
@@ -77,8 +78,8 @@
 				status = "okay";
 
 				port@0 {
-					lcd_output: endpoint {
-						remote-endpoint = <&lvds_encoder_input>;
+					dpi_output: endpoint {
+						remote-endpoint = <&bridge_input>;
 						bus-width = <24>;
 					};
 				};
@@ -1581,7 +1582,7 @@
 		vbus-supply = <&vdd_vbus_usb3>;
 	};
 
-	backlight_lvds: backlight-lvds {
+	backlight: backlight {
 		compatible = "pwm-backlight";
 
 		enable-gpios = <&gpio TEGRA_GPIO(H, 2) GPIO_ACTIVE_HIGH>;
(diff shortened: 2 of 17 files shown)


drivers/mfd/asus-ec-client.c
```diff
@@ -0,0 +1,200 @@
...
+{
+	struct asusec_client_data *priv = dev_get_drvdata(&pdev->dev);
+
+	mfd_remove_devices(&priv->pdev->dev);
+
+	return 0;
+}
+
+static const struct of_device_id asusec_client_match[] = {
+	{ .compatible = "asus,dock-ec" },
```
digetx, line 183: Which other clients than the dock exist? If none, then why this is called "client" and not just "dock" driver.

clamor-s: Hm, in tf701t dock ec is controlled by pad ec and only can be accessed via pad dockram, similar how other subdevs are handled. In t30 devices (except tf600t) dock ec is fully independent i2c device with its own dockram.

drivers/gpu/drm/panel/panel-simple.c
```diff
@@ -2395,6 +2395,31 @@ static const struct panel_desc hannstar_hsd070pww1 = {
...
+	.hsync_len = { 58, 158, 661 },
+	.vactive = { 800, 800, 800 },
+	.vfront_porch = { 1, 1, 10 },
+	.vback_porch = { 1, 1, 10 },
+	.vsync_len = { 1, 21, 203 },
+	.flags = DISPLAY_FLAGS_DE_HIGH,
+};
+
+static const struct panel_desc hannstar_hsd101pww2 = {
+	.timings = &hannstar_hsd101pww2_timing,
```
maximschwalm, line 2412: It would be simpler just to set the timings to `&hannstar_hsd070pww1_timing`

clamor-s: If it is allowed to reuse timings from other panel.

drivers/mfd/asus-ec.c
```diff
@@ -411,6 +436,50 @@ static const struct attribute_group asus_ec_attr_group = {
...
+	if (ret < 0)
+		dev_err(&priv->self->dev, "failed to read dock status: %d\n", ret);
+
+	ret = state[1] & ASUSEC_DOCKED_HYBRID_MASK;
+
+	if (ret) {
+		if (!priv->asusec_docked) {
+			ret = devm_mfd_add_devices(&priv->self->dev, PLATFORM_DEVID_AUTO,
+						   asus_ec_client, ARRAY_SIZE(asus_ec_client),
+						   NULL, 0, NULL);
```
digetx, line 458: You create device that creates other devices, why you can't create those dock devices directly from here?

drivers/mfd/asus-ec.c
```diff
@@ -411,6 +436,50 @@ static const struct attribute_group asus_ec_attr_group = {
...
+			if (ret)
+				dev_err(&priv->self->dev, "failed to add subdevs: %d\n", ret);
+			priv->asusec_docked = true;
+		}
+	} else {
+		if (priv->asusec_docked) {
+			mfd_remove_devices(&priv->self->dev);
```
digetx, line 465: You can't use `devm_` mfd_add_devices and mfd_remove_devices together.

drivers/mfd/asus-ec.c
```diff
@@ -411,6 +436,50 @@ static const struct attribute_group asus_ec_attr_group = {
...
+			ret = asus_ec_init_components(priv, info);
+			if (ret)
+				dev_err(&priv->self->dev, "failed to init subdevs: %d\n", ret);
+
+			priv->asusec_docked = false;
+		}
+	}
+
+	schedule_delayed_work(&priv->poll_work,
+			      msecs_to_jiffies(ASUSEC_DOCKED_CHECK_MSEC));
```
digetx, line 480: Does downstream driver have a similar polling or it uses interrupt?

drivers/mfd/asus-ec.c
clamor-s: Need to check if dock is supported.

drivers/mfd/asus-ec.c
clamor-s: Ok, thanks.

drivers/mfd/asus-ec.c
clamor-s: tf701t uses polling. No interrupt.

drivers/mfd/asus-ec.c
digetx: Why you can't check it here?

clamor-s: I can, but it will turn this unreadable code in total shit.

clamor-s: I did not make a whole separate device. This .c is mostly part of original ec and even shares common config.

digetx: Don't create unnecessary devices, move it to a function. I don't see why you can't make it readable, but you can keep it in a separate file too.

drivers/gpu/drm/panel/panel-simple.c
digetx: it's unlikely that hsd070pww1 will ever be changed, but better to keep it separate.

drivers/mfd/asus-ec.c
clamor-s: Ok, how to bind a dock battery node if pad already has one battery bound?

arch/arm/boot/dts/tegra30-asus-tf201.dts
```diff
@@ -5,6 +5,7 @@
 /* CPU Speedo ID 3, Soc Speedo ID 2, CPU Process: 1, Core Process: 0 */
 
 #include "tegra30-asus-transformer-common.dtsi"
+#include "tegra30-asus-lvds-display.dtsi"
```
clamor-s, line 8: I'm not fan of creating a new dtsi, but this is acceptable. What would you say @digetx?

arch/arm/boot/dts/tegra30-pegatron-chagall.dts
```diff
@@ -1593,65 +1594,8 @@
...
-
-			port@1 {
-				reg = <1>;
-
-				lvds_encoder_output: endpoint {
-					remote-endpoint = <&panel_input>;
-				};
-			};
-		};
+		compatible = "hannstar,hsd101pww2";
```
maximschwalm, line 1598: At least the HannStar panel in Pegatron Chagall has gpios for CABC and Color Engine. What should we do about them?

digetx: I don't know what `Color Engine` is. CABC (content adaptive backlight) is optional, usually you want to have it disabled.

arch/arm/boot/dts/tegra30-asus-tf201.dts
digetx: Okay to me.

drivers/mfd/asus-ec.c
digetx: I don't understand the problem. If the question is about how to differentiate device in the battery driver, then the answer is to use platform data.

arch/arm/boot/dts/tegra30-pegatron-chagall.dts
digetx: You could set them as gpio hogs.

drivers/mfd/asus-ec.c
clamor-s: No, how to bind monitored-battery node for pad and dock?

digetx: Perhaps you need to make monitored-battery to support multiple phandles.

drivers/mfd/asus-ec.c
```diff
@@ -488,6 +554,8 @@ static int asus_ec_remove(struct i2c_client *client)
 {
 	struct asus_ec_data *priv = i2c_get_clientdata(client);
 
+	cancel_delayed_work_sync(&priv->poll_work);
```
clamor-s, line 557: This should be execuded only if polling is working, else it will give error in dmesg. How to add check if priv->poll_work is valid?

drivers/mfd/asus-ec.c
clamor-s: It will lead to changing of core power supply driver and even more serious interference into core functionality.

drivers/mfd/asus-ec-client.c
clamor-s: This compatible can be set to any other or even removed. Battery and charger check for pad-ec to add device anyway.

drivers/mfd/asus-ec.c
digetx: This shouldn't be a problem.

drivers/mfd/asus-ec.c
digetx: Initialize it unconditionally.

clamor-s: Condition is mandatory. If ec is on i2c it mandatiry has irq and request gpio. No polling is needed.

digetx: Move INIT_DELAYED_WORK out of condition.

clamor-s: Reasonable, but maybe better to add bool check into main struct to avoid even init of work for non dedicated devices?

drivers/mfd/asus-ec.c
clamor-s: Does this effort worth time? It will affect all devices that use monitored-battery node.

digetx: You should add new a function for that, other devices will be unaffected.

drivers/mfd/asus-ec.c
digetx: That is unnecessary.

drivers/mfd/asus-ec.c
clamor-s: I will not interfare with core power driver if I can avoid it. I will drop this commit in update.

digetx: That may be inevitable if h/w description is correct and both batteries are indeed connected to the same EC. On the other hand, I'm not sure why you need to use monitored battery at all since you can hadrcode all battery parameters in the driver per device model.

Technically, I could hardcode battery params for A500, but they are not used anyways and so it's unimportant. In case of Nexus 7 I removed monitored battery from the battery controller recently, controller knows all parameters anyways. The charger of Nexus 7 is a different story because it's a common h/w used by different tablets, each device uses own battery and charger wants to know some of battery parameters.

clamor-s:
@digetx when/if you merge this PR, please add Max as a co-developer into tf700t commit and remove WIP from chagall dts commit. 

P. S. I will work with tf701t dock a bit more. It is not that important.

drivers/power/supply/asus-ec-charger.c
```diff
@@ -132,6 +134,7 @@ static int asusec_charger_probe(struct platform_device *pdev)
 	mutex_init(&priv->charger_lock);
 
 	priv->ec = ec;
+	priv->charger_addr = pdata->charger_addr;
 	priv->charger_data_ts = jiffies - 1;
 	priv->last_state = asusec_charger_callback(priv);
```
digetx, line 147: Note that you can specify the fixed power supply for the battery in the code, like I did if for the surface-rt charger https://github.com/grate-driver/linux/blob/master/drivers/power/supply/surface-rt-ec.c#L350

digetx:
Thanks!

drivers/power/supply/asus-ec-charger.c
clamor-s: Ok, I will think about this since it requires a bit of accuracy with proper routing.

clamor-s: It should go to both pad and dock battery I suppose.

PR #69, opened 2021-08-31
Add minimal defconfig and pmos config fragment

jenneron:
grate_defconfig devices support:
 - acer-picasso;
 - surface rt (except bluetooth, we still don't have firmware);
 - transformers (tf300t not tested);
 - pegatron-chagall (not tested);
 - grouper / tilapia (not tested);
 
 To use that you need to add at least panel and FS related modules to initramfs and generate new initramfs when updating kernel.
 
 pmos.config besides generic needed options:
  - built-in stuff needed for cryptsetup;
  - anbox support (is broken on armv7, but we hope it will be fixed);
  - nftables;
  - containers (docker, lxc);
  - zram.

Some references for contributors:
- grate_defconfig should not be distro-specific and feel free to update it with your device, especially if you want to enable it in pmOS, but use modules for device-specific things when possible;
- to get pmOS-capable config use make `grate_defconfig pmos.config`;
- to update grate_defconfig use something like `make grate_defconfig && make menuconfig && make savedefconfig && cp defconfig arch/arm/configs/grate_defconfig`. For pmOS with envkernel use `make grate_defconfig && make menuconfig && make savedefconfig && cp .output/defconfig arch/arm/configs/grate_defconfig && make grate_defconfig pmos.config`
- keep on mind that for upstream you will have to use tegra_defconfig and multi_v7_defconfig. This defconfig is for WIP devices and to support devices like grouper which have bootloader limitations.

`make grate_defconfig pmos.config` gives 3.1M zImage, just `make grate_defconfig` should give smaller one. As grouper accepts < 3.5M we still have reserve to support more devices.


okias:
I have prepared the CI, which test it just by downloading config from pmOS pmaports, wouldn't be that enough? 
I suppose this isn't meant for upstreaming, so it just adds more commits a top of -next to maintain.

jenneron:
> I have prepared the CI, which test it just by downloading config from pmOS pmaports, wouldn't be that enough?

It makes things easier to maintain in pmOS and easier to test locally.
We can update these configs here once new device features are comming. So if someone adds a new device with new configs they may not care about updating kernel config in pmaports. I (or someone else) will just do `make grate_defconfig pmos.config` for pmaports and new device is suppored.
For contributors it's convenient to do `make grate_defconfig` or `make grate_defconfig pmos.config` because `make tegra_defconfig` won't fit to boot partition on most devices and besides that it won't boot on pmOS due to missing dm-crypt stuff.

jenneron:
> I suppose this isn't meant for upstreaming, so it just adds more commits a top of -next to maintain.

I will maintain pmos.config myself, so you may not care about it. Future grate_defconfig changes should be same to tegra_defconfig changes except WIP devices. Should not be too hard to maintain

digetx:
Thanks!

PR #71, opened 2021-09-03
FIXUP: CI: add pmOS-grate kernel configs

okias:
grate_defconfig + pmos.config (so it builds nightly pmOS kernel)


digetx:
Thanks!

PR #75, opened 2021-09-12
transformers: chagall: update dts bindings

clamor-s:


clamor-s:
@okias assume this would be enough

digetx:
Thanks!

PR #76, opened 2021-09-12
Isa1200 rename

okias:
rename + few fixups for glide, added simple yaml definition...

oneplus used also isa1200, seems it has nice docs to incorporate later: https://searchcode.com/file/260805862/Documentation/devicetree/bindings/misc/isa1200.txt/


digetx:
Thanks!

PR #77, opened 2021-09-13
dt-bindigs: arm: tegra20: Acer A500 Picasso correct al8975 compatible

okias:
"ak,ak8975" is not recognized by dt-bindings, "ak8975" is deprecated so
let's use supported "asahi-kasei,ak8975"


digetx:
Thanks!

PR #79, opened 2021-09-17
dtbs_check fixups

okias:
Multiple small nitpicks to reduce warnings printed in dtbs_check

### DIFF
--- a/arch/arm/boot/dts/tegra20-acer-a500-picasso.dts
+++ b/arch/arm/boot/dts/tegra20-acer-a500-picasso.dts
@@ -825,7 +825,7 @@
 	};
 
 	/* PMIC has a built-in 32KHz oscillator which is used by PMC */
-	clk32k_in: clock@0 {
+	clk32k_in: clock-32k-in {
 		compatible = "fixed-clock";
 		#clock-cells = <0>;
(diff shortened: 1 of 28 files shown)


arch/arm/boot/dts/tegra20-acer-a500-picasso.dts
```diff
@@ -837,7 +837,7 @@
 	 * oscillator is used as a reference clock-source by the
 	 * Azurewave WiFi/BT module.
 	 */
-	rtc_32k_wifi: clock@1 {
+	rtc_32k_wifi: clock-32k-wifi {
```
digetx, line 840: Could you please make a single patch which makes all these renamings? Regulator nodes should also produce the warnings, AFAICS. Please rename them in a similar way `s/vdd_5v0_sys: regulator@0/vdd_5v0_sys: regulator-5v/`.

Please fix all other mainlined ARM32 tegra boards too (ventana, cardhu and etc), to make this patch upstreamable.

okias: Ok, I'll do one patch for all mainline boards and remove it from this MR. Thanks

digetx:
Thanks!

PR #83, opened 2021-09-27
Transformers

clamor-s:
- add missing backlight regulator (testing with u-boot found this issue);
- fix bt compatible fields and irq type


digetx:
Thanks!

PR #88, opened 2021-09-30
multiple DTS fixes + CI adjustment

okias:


arch/arm/boot/dts/tegra30-asus-tf700t.dts
```diff
@@ -140,7 +140,7 @@
 				 * Panasonic VVX10F004B00 or HYDIS HV101WU1-1E1
 				 * LCD SuperIPS+ Full HD panel.
 				 */
-				display-panel@1 {
+				panel@1 {
```
digetx: Does this fix a warning? A500, N7 and all other transformers use `display-panel` name, they should also produce the warning then, shouldn't they?

okias: doesn't fix directly any warning, but since nodes containing display panel are usually named panel@ I renamed the node to match that schematic. I could keep it as display-panel and alter yaml Doc to reflect that

digetx: Ok, no need to change it.

digetx:
Thanks!

PR #89, opened 2021-10-01
Draft: tegra-ehci txt -> yaml

okias:
~ 900 warnings down.


Documentation/devicetree/bindings/usb/generic-ehci.yaml
```diff
@@ -148,12 +191,23 @@ properties:
...
+    description: |
+      Some of the Tegra20 USB ports, need reset twice due to hardware issues.
+
 required:
   - compatible
   - reg
   - interrupts
 
-additionalProperties: false
+unevaluatedProperties: false
```
digetx, line 210: The USB binding should require a separate yaml, there are to many tegra-specific things which don't deserve to be placed in generic binding.

There is unfinished binding from @thierryreding https://patchwork.ozlabs.org/project/linux-tegra/list/?series=249961 Maybe you could try to address those comments and send v2, if Thierry not objecting.

PR #94, opened 2021-10-07
fixup! ARM: grate_defconfig: Add defconfig (v2)

jenneron:
Enable preemption which was disabled with my latest savedefconfig and pull request.

I'm not sure which one should be enabled, i enabled a `CONFIG_PREEMPT_VOLUNTARY_BEHAVIOUR` one for now. Let me know if i should enable `CONFIG_PREEMPT_BEHAVIOUR` instead. If so, i will update this PR

tegra_defconfig may should be updated as well


digetx:
Thanks!

PR #97, opened 2021-10-09
fixup! ARM: grate_defconfig: Add defconfig (v3)

jenneron:
Add support for glide and enable `CONFIG_INPUT_UINPUT` needed for fbkeyboard

for now it is draft, once it is tested on glide i will mark it as ready


digetx:
Thanks!

PR #105, opened 2021-10-23
dt-schema: Nvidia tegra audio (3 similar patches)

okias:


Documentation/devicetree/bindings/sound/nvidia,tegra-audio.yaml
```diff
@@ -0,0 +1,205 @@
...
+          * Headset Stereophone
+          * Int Spk
+          * Headset Mic
+          * Digital Mic
+        MAXX98090:
+          * Headphones
+          * SPeakers
+          * Mic Jack
+          * Int Mic
+        RT5631:
```
digetx, line 97: RT5631 shouldn't be in this patch.

PR #111, opened 2021-11-04
chagall: fix regulators in dts

clamor-s:
minor change in chagall power set according to downstream

### DIFF
--- a/arch/arm/boot/dts/tegra30-pegatron-chagall.dts
+++ b/arch/arm/boot/dts/tegra30-pegatron-chagall.dts
@@ -1328,8 +1328,8 @@
 
 				ldo5 {
 					regulator-name = "vdd_1v3_cam_isp";
-					regulator-min-microvolt = <1000000>;
-					regulator-max-microvolt = <1000000>;
+					regulator-min-microvolt = <1300000>;
+					regulator-max-microvolt = <1300000>;
 				};
 
 				ldo6 {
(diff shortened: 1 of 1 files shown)


digetx:
Thanks!

arch/arm/boot/dts/tegra30-pegatron-chagall.dts
```diff
@@ -1416,6 +1416,17 @@
 		vin-supply = <&vdd_3v3_sys>;
 	};
 
+	vdd_5v0_bl: regulator-bl {
+		compatible = "regulator-fixed";
+		regulator-name = "vdd_5v0_bl";
+		regulator-min-microvolt = <5000000>;
+		regulator-max-microvolt = <5000000>;
+		regulator-boot-on;
+		gpio = <&gpio TEGRA_GPIO(C, 6) GPIO_ACTIVE_HIGH>;
```
digetx, line 1425: Could you please give a reference to the downstream code?

clamor-s: https://github.com/sonntam/android_kernel_fujitsu_mx532/blob/master/arch/arm/mach-tegra/board-chagall-power.c#L574

digetx: Thanks

PR #115, opened 2021-11-25
Transformer & Chagall updates

maximschwalm:
Add some FIXME tags and fix up device-tree commits

### DIFF
--- a/arch/arm/boot/dts/tegra20-asus-tf101.dts
+++ b/arch/arm/boot/dts/tegra20-asus-tf101.dts
@@ -405,7 +405,7 @@
 		compatible = "nvidia,tegra20-hsuart";
 		status = "okay";
 
-		/* Azurewave AW-NH665 BCM4329B1 */
+		/* Azurewave AW-NH615 BCM4329B1 */
 		bluetooth {
 			compatible = "brcm,bcm4329-bt";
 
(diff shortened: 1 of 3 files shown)


digetx:
Are those tags really useful? Won't be better to use issue tracker instead of the comments in the code?

maximschwalm:
> Are those tags really useful? Won't be better to use issue tracker instead of the comments in the code?

Yeah, it might be good to use the issue tracker as well. The comments just make the issues more noticable and documentation doesn't hurt IMO. I found other examples like [this](https://github.com/grate-driver/linux/blob/master/arch/arm64/boot/dts/qcom/msm8916-samsung-serranove.dts#L290) one.

digetx:
Well, those comments about downstream kernel aren't appropriate for the upstream. Do you want to have those comments in grate kernel only? The old MMC comment was okay to me.

maximschwalm:
> Well, those comments about downstream kernel aren't appropriate for the upstream. Do you want to have those comments in grate kernel only? The old MMC comment was okay to me.

SDR104 mode works though just not with full clock speed. Is the comment about VIO better now?

digetx:
The VIO comment now is better. I would also replace `Full clock` with `Full 208MHz clock rate` .

digetx:
I'll edit the comments after applying, thanks!

PR #116, opened 2021-12-07
Chagall & Transformer fixups

maximschwalm:
- Add some more device-tree comments

- Try to improve commit messages

- Add me as a co-developer of the TF300TG

### DIFF
--- a/arch/arm/boot/dts/tegra30-asus-tf201.dts
+++ b/arch/arm/boot/dts/tegra30-asus-tf201.dts
@@ -138,6 +138,7 @@
 		vcc-supply = <&vdd_3v3_sys>;
 	};
 
+	/* FIXME: Sound only works, if booted from downstream kernel */
 	sound {
 		compatible = "asus,tegra-audio-rt5631-tf201",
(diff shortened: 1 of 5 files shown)


digetx:
References to downstream kernel aren't useful for upstream, I'll remove them. I'll merge this, although I already asked Thierry to pull v4 couple days ago. Will see if it's not too late for these minor changes. Thanks!

digetx:
Ah, I see now why we had problems with the rebase previously. I'm doing the "rebase" variant of the merge and some patches are getting discarded in this case. I'll apply the missing patches manually.

digetx:
I see now that those downstream references are made only for the `WIP` patch that isn't going to upstream, so I won't remove them.

digetx:
I'll move the Chagall comment about the sound into the WIP patch.

maximschwalm:
> I'll move the Chagall comment about the sound into the WIP patch.

The same comment was also added to the TF300T DT.

digetx:
Oh, well. Please don't forget to remove it once sound will work.

PR #117, opened 2021-12-12
Tegra complex binding rework

okias:
WM8994 complex Connectors and Pins seems to be completly wrong, rest should be ok


digetx:
Thanks!

PR #119, opened 2022-01-21
Asus Transformers patchset

clamor-s:


arch/arm/boot/dts/tegra20-asus-tf101.dts
```diff
@@ -450,11 +450,15 @@
 
 	serial@70006040 {
 		compatible = "nvidia,tegra20-hsuart";
-		/* GPS BCM4751 */
+		/delete-property/ reg-shift;
+		status = "okay";
+
+		/* Broadcom GPS BCM47511 */
```
maximschwalm, line 456: BCM4751 was certainly correct

clamor-s: Ok

arch/arm/boot/dts/tegra30-asus-transformer-common.dtsi
```diff
@@ -1469,6 +1471,8 @@
 			interrupt-parent = <&gpio>;
 			interrupts = <TEGRA_GPIO(O, 4) IRQ_TYPE_LEVEL_HIGH>;
 			interrupt-names = "host-wake";
+
+			reset-gpios = <&gpio TEGRA_GPIO(D, 3) GPIO_ACTIVE_LOW>;
```
digetx, line 1475: Could you please explain how this was tested? This "reset-gpios" isn't used by the kernel's brcmfmac driver, AFAICS.

1. I see that Chagall and Ouya use D3 in the `wifi_pwrseq`, it's likely to be correct.
2. You need to check whether toggling D4 makes a real difference (WiFi doesn't work without enabling it) because this gpio may be physically unconnected on Transformers, similarly to other Tegra devices where reset gpio is connected to both RST and PWR-enb pins of BCM WiFi (Acer A500, Nexus7 and etc).
3. If D4 really works, then it should be specified as `vmmc-supply` and D3 should go to the pwrseq.

clamor-s: Then why this property is in yaml? There was tests a while ago. wifi_pwrseq uses D4, and vmmc-supply is D0. D3 as I remember did not make any difference, although downstream uses it as reset gpio (https://github.com/clamor-s/morpheus_kernel_asus_tf201/blob/kernel-rework/arch/arm/mach-tegra/board-cardhu-sdhci.c#L38). This is to your choice.

digetx: Downstream kernel specifies GPIOs for all board variants, some of those GPIOs may do nothing depending on a board. If D3 does nothing on Transformers, then we shouldn't use it.

I'll merge this PR and drop this patch on rebase.

clamor-s: Sure

arch/arm/boot/dts/tegra30-asus-transformer-common.dtsi
```diff
@@ -12,9 +12,9 @@
 	chassis-type = "convertible";
 
 	aliases {
-		mmc0 = "/mmc@78000600"; /* eMMC */
-		mmc1 = "/mmc@78000000"; /* uSD slot */
-		mmc2 = "/mmc@78000400"; /* WiFi */
+		mmc0 = &sdmmc4; /* eMMC */
+		mmc1 = &sdmmc1; /* uSD slot */
+		mmc2 = &sdmmc3; /* WiFi */
```
digetx, line 17: All such label changes should be squashed into a single patch and applied to all boards uniformly. I'll keep this patch in grate-kernel for now, but it will need to be extended if we'll want to merge it into upstream. You may continue working on it with further patches.

digetx:
Thanks!

arch/arm/boot/dts/tegra30-asus-transformer-common.dtsi
clamor-s: Only tf has this change, tf101 and chagall have labels, but t30 transformers for some reasons did not fit. Idk why.

clamor-s: It was in this way when sent to upstream, but reviewers decided to change it. I wonder why?

digetx: @thierryreding didn't want to apply patch that added labels to dtsi because some labels were unused.

PR #120, opened 2022-03-08
Chagall & Transformer EC updates

maximschwalm:


drivers/leds/leds-asus-ec.c
```diff
@@ -14,24 +14,24 @@
...
  *  auto: brightness == 0
  *  bit 0: blink / charger on
- *  bit 1: amber on
+ *  bit 1: orange on
  *  bit 2: green on
  */
 
 #define ASUSEC_CTL_LED_BLINK		BIT_ULL(40)
-#define ASUSEC_CTL_LED_AMBER		BIT_ULL(41)
+#define ASUSEC_CTL_LED_ORANGE		BIT_ULL(41)
```
digetx, line 22: The amber was correct. It's the A500 LED driver which should be corrected :)

digetx:
Thanks!

PR #123, opened 2022-05-18
Add support for Asus Portable AiO P1801-T

maximschwalm:


digetx:
Thanks @maximschwalm. I'm going to review this over the weekend.

drivers/usb/misc/usb_pwrseq.c
```diff
@@ -0,0 +1,59 @@
...
+		return -ENOMEM;
+
+	data->reset_gpio = devm_gpiod_get_optional(dev,"reset", GPIOD_OUT_HIGH);
+	if (IS_ERR(data->reset_gpio))
+		return dev_err_probe(dev, PTR_ERR(data->reset_gpio),
+					 "failed to get reset GPIO\n");
+
+	if (data->reset_gpio) {
+		msleep(1);
+		gpiod_set_value(data->reset_gpio, 0);
```
digetx, line 31: Usually reset should de-asserted after power is stable.

arch/arm/boot/dts/tegra30-asus-p1801-t.dts
```diff
@@ -0,0 +1,2096 @@
...
+			label = "Volume Down";
+			gpios = <&gpio TEGRA_GPIO(Q, 3) GPIO_ACTIVE_LOW>;
+			linux,code = <KEY_VOLUMEDOWN>;
+			debounce-interval = <10>;
+			wakeup-event-action = <EV_ACT_ASSERTED>;
+			wakeup-source;
+		};
+
+		mode {
+			label = "Mode";
```
digetx, line 134: What is the "Mode"?

clamor-s: https://github.com/clamor-s/morpheus_kernel_asus_tf201/blob/asus/arch/arm/mach-tegra/board-cardhu-kbc.c#L71

clamor-s: Not quite sure but most likely it is pad/dock cpu switch

drivers/usb/misc/usb_pwrseq.c
maximschwalm: The sequence is from [here](https://github.com/clamor-s/morpheus_kernel_asus_tf201/blob/asus/arch/arm/mach-tegra/board-cardhu.c#L1239-L1252). Should I still move it around? BTW, the reset sequence of the sis_i2c driver seems to be kind of similiar but it goes from low to high to low. Is it more correct? Besides, the downstream kernel has a "tp_vendor" GPIO. Is there a use case for it? Thanks.

arch/arm/boot/dts/tegra30-asus-p1801-t.dts
maximschwalm: The tablet can be switched to PC mode by a press of this button. The display input switches to the PC docking station and the USB devices disconnect from the tablet. The HPD state changes as a result. When the tablet is detached nothing happens though. @hedmo could tell more about it.

arch/arm/boot/dts/tegra30-asus-p1801-t.dts
```diff
@@ -0,0 +1,2107 @@
...
+
+    connector {
+		compatible = "hdmi-connector";
+		label = "HDMI";
+		type = "a";
+
+		/* low: tablet, high: dock */
+		hpd-gpios = <&gpio TEGRA_GPIO(H, 4) GPIO_ACTIVE_LOW>;
+		ddc-i2c-bus = <&i2c4>;
+		ddc-en-gpios = <&gpio TEGRA_GPIO(H, 3) GPIO_ACTIVE_HIGH>;
```
maximschwalm, line 85: I realized that the display-connector driver has no support for this GPIO, so this doesn't do anything. It might also be an enable GPIO for the bridge, but when I set it as such I don't get display output:
`drm drm: [drm] Cannot find any crtc or sizes`
Then it falls back to 1024x768 mode and I see just the backlight.

clamor-s:
@digetx may you remove transformers branch? Those changes are in mainline already

digetx:
Removed

arch/arm/boot/dts/tegra30-asus-p1801-t.dts
digetx: @maximschwalm Will you you be able to clarify that GPIO?

arch/arm/boot/dts/tegra30-asus-p1801-t.dts
digetx: What about:

```diff
- mode {
+ docking-station-mode-switch {
```

drivers/usb/misc/usb_pwrseq.c
```diff
@@ -0,0 +1,59 @@
...
+static int usb_pwrseq_probe(struct platform_device *pdev)
+{
+	struct usb_pwrseq_data *data;
+	struct device *dev = &pdev->dev;
+
+	data = devm_kzalloc(dev, sizeof(*data), GFP_KERNEL);
+	if (!data)
+		return -ENOMEM;
+
+	data->reset_gpio = devm_gpiod_get_optional(dev,"reset", GPIOD_OUT_HIGH);
```
digetx, line 24: Nit: whitespace is missing after `(dev,`

arch/arm/boot/dts/tegra30-asus-p1801-t.dts
maximschwalm: The GPIO is referred to as EN_VDD_BL also scalar_enable in the P1802 kernel.  If I manually toggle the GPIO via sysfs, I can power on/off the display.

hedmo: this sounds like the same as using the magic button . if i press it when docked i get the text : "dock mode" and "pad mode" when pressed again. if the pad is undocked the button wont work. when you toggle  it you maybe putting the pad i  "docked mode"

drivers/usb/misc/usb_pwrseq.c
digetx: @maximschwalm Are you going to push update anytime soon? If not, then I can merge what we have now and we'll continue with a new PRs.

maximschwalm: I applied your suggestions except for changing the power sequence of the USB device. Is it good to go then?

digetx: Yes, thanks!

PR #126, opened 2022-08-01
P1801-T & grate_defconfig updates

maximschwalm:
Cc @jenneron


arch/arm/boot/dts/tegra30-asus-p1801-t.dts
```diff
@@ -2096,9 +2085,35 @@
...
+			reset-gpios = <&gpio TEGRA_GPIO(H, 6) GPIO_ACTIVE_LOW>;
+			power-gpios = <&gpio TEGRA_GPIO(S, 0) GPIO_ACTIVE_HIGH>;
+		};
+
+		/* Chicony USB2.0 HD UVC WebCam */
+		device@2 {
+			compatible = "usb4f2,b354";
+			reg = <2>;
+			power-gpios = <&gpio TEGRA_GPIO(BB, 4) GPIO_ACTIVE_HIGH>;
+		};
```
digetx, line 2113: Could we use gpio hogs for these USB devices?

In general we shouldn't amend changes that aren't upstreamable right now, such changes should be kept in a separate patches on top of the more ready patches. Hence there should be the basic tegra30-asus-p1801-t.dts that we can send to upstream today and everything else should be made on top of it.

digetx: We can improve this later on.

digetx:
Thanks

PR #127, opened 2022-08-05
Update sound routing and convert tegra gpio 

clamor-s:
- gpio-tegra driver needs a review, if remove of tegra_gpio_irq_set_affinity does not brake anything;
- sound was reworked but it misses key part, headset detection. Asus made dumb stuff, Transformers have hp detect, not headset.
PW2 = low   PX2 = low --- default
PW2 = high  PX2 = low --- headset
PW2 = high  PX2 = high --- headphones
tegra_asoc_machine needs a workarround :( it may work for grouper as well btw.

### DIFF
--- a/drivers/gpio/gpio-tegra.c
+++ b/drivers/gpio/gpio-tegra.c
@@ -18,6 +18,7 @@
 #include <linux/of_device.h>
 #include <linux/platform_device.h>
 #include <linux/module.h>
+#include <linux/seq_file.h>
 #include <linux/irqdomain.h>
 #include <linux/irqchip/chained_irq.h>
 #include <linux/pinctrl/consumer.h>
@@ -94,7 +95,6 @@ struct tegra_gpio_info {
 	struct tegra_gpio_bank			*bank_info;
 	const struct tegra_gpio_soc_config	*soc;
 	struct gpio_chip			gc;
-	struct irq_chip				ic;
 	u32					bank_count;
 	unsigned int				*irqs;
 };
@@ -288,6 +288,7 @@ static void tegra_gpio_irq_mask(struct irq_data *d)
 	unsigned int gpio = d->hwirq;
 
 	tegra_gpio_mask_write(tgi, GPIO_MSK_INT_ENB(tgi, gpio), gpio, 0);
+	gpiochip_disable_irq(chip, gpio);
 }
 
 static void tegra_gpio_irq_unmask(struct irq_data *d)
@@ -296,6 +297,7 @@ static void tegra_gpio_irq_unmask(struct irq_data *d)
 	struct tegra_gpio_info *tgi = gpiochip_get_data(chip);
 	unsigned int gpio = d->hwirq;
 
+	gpiochip_enable_irq(chip, gpio);
 	tegra_gpio_mask_write(tgi, GPIO_MSK_INT_ENB(tgi, gpio), gpio, 1);
 }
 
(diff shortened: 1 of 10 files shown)


drivers/gpio/gpio-tegra.c
```diff
@@ -598,10 +592,31 @@ static void tegra_gpio_irq_release_resources(struct irq_data *d)
...
+}
+
+static const struct irq_chip tegra_gpio_irq_chip = {
+	.irq_shutdown		= tegra_gpio_irq_shutdown,
+	.irq_ack		= tegra_gpio_irq_ack,
+	.irq_mask		= tegra_gpio_irq_mask,
+	.irq_unmask		= tegra_gpio_irq_unmask,
+	.irq_set_type		= tegra_gpio_irq_set_type,
+#ifdef CONFIG_PM_SLEEP
+	.irq_set_wake		= tegra_gpio_irq_set_wake,
```
maximschwalm, line 609: Here you may use the pm_sleep_ptr() macro

clamor-s: Sure, when digetx confirms that patch is performed correctly.

drivers/gpio/gpio-tegra.c
```diff
@@ -569,16 +573,6 @@ static int tegra_gpio_irq_set_wake(struct irq_data *d, unsigned int enable)
 }
 #endif
 
-static int tegra_gpio_irq_set_affinity(struct irq_data *data,
-				       const struct cpumask *dest,
-				       bool force)
-{
-	if (data->parent_data)
-		return irq_chip_set_affinity_parent(data, dest, force);
```
digetx, line 577: This function should be removed. You'll need to add a separate `struct irq_chip` for T210.

digetx: *shouldn't

clamor-s: so patch is correct?

clamor-s: if function should not be removed, then how to assign it properly?

digetx:
```c
np = of_find_matching_node(NULL, tegra_pmc_of_match);
if (np) {
...
	gpio_irq_chip_set_chip(irq, &tegra210_gpio_irq_chip);
} else {
	gpio_irq_chip_set_chip(irq, &tegra_gpio_irq_chip);
}
```

clamor-s: hm, alright do not go anywhere

clamor-s: only t210 uses affinity or only t210 DOES NOT use it

clamor-s: @digetx check now please

drivers/gpio/gpio-tegra.c
```diff
@@ -598,10 +602,43 @@ static void tegra_gpio_irq_release_resources(struct irq_data *d)
...
+	seq_printf(s, dev_name(chip->parent));
+}
+
+static const struct irq_chip tegra_gpio_irq_chip = {
+	.irq_shutdown		= tegra_gpio_irq_shutdown,
+	.irq_ack		= tegra_gpio_irq_ack,
+	.irq_mask		= tegra_gpio_irq_mask,
+	.irq_unmask		= tegra_gpio_irq_unmask,
+	.irq_set_type		= tegra_gpio_irq_set_type,
+	.irq_set_wake		= pm_sleep_ptr(&tegra_gpio_irq_set_wake),
```
digetx, line 618: `&` not needed, function itself is a pointer already

drivers/gpio/gpio-tegra.c
```diff
@@ -752,8 +776,9 @@ static int tegra_gpio_probe(struct platform_device *pdev)
 		if (!irq->parent_domain)
 			return -EPROBE_DEFER;
 
-		tgi->ic.irq_set_affinity = tegra_gpio_irq_set_affinity;
-	}
+		gpio_irq_chip_set_chip(irq, &tegra210_gpio_irq_chip);
+	} else 
+		gpio_irq_chip_set_chip(irq, &tegra_gpio_irq_chip);
```
digetx, line 781: https://www.kernel.org/doc/html/latest/process/coding-style.html#placing-braces-and-spaces

drivers/gpio/gpio-tegra.c
```diff
@@ -286,16 +286,20 @@ static void tegra_gpio_irq_mask(struct irq_data *d)
 	struct gpio_chip *chip = irq_data_get_irq_chip_data(d);
 	struct tegra_gpio_info *tgi = gpiochip_get_data(chip);
 	unsigned int gpio = d->hwirq;
+	int pin = irqd_to_hwirq(d);
```
digetx, line 289: `irqd_to_hwirq(d) = d->hwirq`, no need for the extra variable

clamor-s:
[SQUASH] ASoC: tegra: Support RT5631 by machine driver 930e9a4d8f0131fba3133c0f92f5b94bf8c0acb4 -- includes mic jack detection fix

I can squash all dts changes for transformers into one commit, may be a correct solution.

gpio: tegra: Convert to immutable irq chip b4b86583ab3630ca1a849388dc06de6ff24ffb14 -- fixed

clamor-s:
I have squashed dts changes into one commit. There are also small patches into dsp (set it tristate) and asusec-kbc (change camera key to print key). I will amend those once this patch is approved.

sound/soc/tegra/tegra_asoc_machine.c
```diff
@@ -63,6 +75,7 @@ static struct snd_soc_jack_gpio tegra_machine_mic_jack_gpio = {
 	.name = "Mic detection",
 	.report = SND_JACK_MICROPHONE,
 	.debounce_time = 150,
+	.jack_status_check = headset_check,
```
digetx, line 78: This will break other add_mic_jack users.

clamor-s: ok, I added dts flag to show that device has mic-det gpio has to be used only if hp-det gpio is active. This should not be codec dependent, both rt5631 and wm8903 transformers have this, it is device dependent.

sound/soc/tegra/tegra_asoc_machine.c
```diff
@@ -189,8 +200,12 @@ int tegra_asoc_machine_init(struct snd_soc_pcm_runtime *rtd)
 			return err;
 		}
 
+		tegra_machine_mic_jack_gpio.data = machine;
 		tegra_machine_mic_jack_gpio.desc = machine->gpiod_mic_det;
 
+		if (machine->bound_mic_jack)
```
digetx, line 206: This `machine->bound_mic_jack` is unnecessary, use `of_property_read_bool` directly.

The `nvidia,bound-mic-jack` doesn't sound good to me, it doesn't explain what is bounded. Something like `nvidia,bounded-mic-hp-det` or `nvidia,coupled-mic-hp-det` should be better, IMO.

digetx: You'll also need to update the DT binding.

clamor-s: @digetx updated

digetx: You'll need to document the new `nvidia,coupled-mic-hp-det` property in the DT binding.

digetx:
Thanks

sound/soc/tegra/tegra_asoc_machine.c
clamor-s: I will refit and incorporate those patches. Are you ok with that?

PR #128, opened 2022-08-06
config changes

jenneron:


kernel/configs/pmos.config
```diff
@@ -53,7 +53,7 @@ CONFIG_ASHMEM=y
 CONFIG_ANDROID=y
```
maximschwalm, line 53: Should we remove this deprecated option and enable PM_USERSPACE_AUTOSLEEP instead?

digetx: We still have CONFIG_ANDROID in grate kernel for legacy ramconsole, but we don't really need that kconfig option and may remove it.

PM_USERSPACE_AUTOSLEEP should make system unusable, it's intended for Android only.

maximschwalm: It's no longer possible to build legacy ramconsole.

digetx: Indeed, I missed that the Kconfig needed more changes after last rebase.

digetx: I'll fix the ramconsole on the next rebase.

digetx:
Thanks

PR #137, opened 2023-08-01
dt-bindings: arm: tegra: Add new devices

maximschwalm:
Add compatibles for the following devices:

- Asus Portable AiO P1801-T
- Asus VivoTab RT TF600T
- LG Optimus Vu P895


digetx:
thanks

Issue #108, opened 2021-10-27
Surface: wm8962 missing supplies


okias:
```
/home/runner/work/linux/linux/arch/arm/boot/dts/tegra30-microsoft-surface-rt.dt.yaml: wm8962@1a: 'AVDD-supply' is a required property
	From schema: /home/runner/work/linux/linux/Documentation/devicetree/bindings/sound/wlf,wm8962.yaml
/home/runner/work/linux/linux/arch/arm/boot/dts/tegra30-microsoft-surface-rt.dt.yaml: wm8962@1a: 'CPVDD-supply' is a required property
	From schema: /home/runner/work/linux/linux/Documentation/devicetree/bindings/sound/wlf,wm8962.yaml
/home/runner/work/linux/linux/arch/arm/boot/dts/tegra30-microsoft-surface-rt.dt.yaml: wm8962@1a: 'DBVDD-supply' is a required property
	From schema: /home/runner/work/linux/linux/Documentation/devicetree/bindings/sound/wlf,wm8962.yaml
/home/runner/work/linux/linux/arch/arm/boot/dts/tegra30-microsoft-surface-rt.dt.yaml: wm8962@1a: 'DCVDD-supply' is a required property
	From schema: /home/runner/work/linux/linux/Documentation/devicetree/bindings/sound/wlf,wm8962.yaml
/home/runner/work/linux/linux/arch/arm/boot/dts/tegra30-microsoft-surface-rt.dt.yaml: wm8962@1a: 'MICVDD-supply' is a required property
	From schema: /home/runner/work/linux/linux/Documentation/devicetree/bindings/sound/wlf,wm8962.yaml
/home/runner/work/linux/linux/arch/arm/boot/dts/tegra30-microsoft-surface-rt.dt.yaml: wm8962@1a: 'PLLVDD-supply' is a required property
	From schema: /home/runner/work/linux/linux/Documentation/devicetree/bindings/sound/wlf,wm8962.yaml
/home/runner/work/linux/linux/arch/arm/boot/dts/tegra30-microsoft-surface-rt.dt.yaml: wm8962@1a: 'SPKVDD1-supply' is a required property
	From schema: /home/runner/work/linux/linux/Documentation/devicetree/bindings/sound/wlf,wm8962.yaml
/home/runner/work/linux/linux/arch/arm/boot/dts/tegra30-microsoft-surface-rt.dt.yaml: wm8962@1a: 'SPKVDD2-supply' is a required property
[...]

PR #140, opened 2026-06-20
ARM: dts: tegra30-microsoft-surface-rt: Fix WM8962 SPKVDD supply

blakegordon:
## Summary

Fix silent audio on Microsoft Surface RT by wiring the WM8962 Class-D speaker
amplifier supplies (SPKVDD1/2) to the board 5 V rail instead of 1.8 V.

## Problem

On postmarketOS with `linux-postmarketos-grate`, the WM8962 codec probes
successfully and ALSA playback appears to work, but no sound is produced from
the internal speakers or headphones. During playback, mixer routes and Class-D
enable bits look correct in software.

`dmesg` reports on every playback attempt:

```
wm8962: ASoC error (-16) at register 0x30 (ADDITIONAL_CONTROL_4)
wm8962: DC servo timed out
```

## Root cause

Per the WM8962 datasheet, SPKVDD must be in the **2.5 V–5.5 V** range for the
Class-D speaker outputs (SPKOUTL/R). The device tree currently assigns both
`SPKVDD1-supply` and `SPKVDD2-supply` to `vdd_1v8_sys` (1.8 V), which is below
the minimum operating voltage for the speaker amplifier.

Digital and analog core supplies correctly remain on 1.8 V; only the speaker
amplifier rails need the higher voltage.

## Fix

Point `SPKVDD1-supply` and `SPKVDD2-supply` at `vdd_5v0_sys`, which is already
modeled in the device tree and used by other board peripherals (backlight, USB
VBUS, etc.).

The EFI variant (`tegra30-microsoft-surface-rt-efi.dts`) includes this file, so
no separate change is required there.

## Testing

- [ ] Rebuild/install kernel with this DTS change on Surface RT
- [ ] Verify internal speaker output with `speaker-test -Dhw:1,0`
- [ ] Verify headphone output
- [ ] Confirm `dmesg` no longer reports WM8962 register `0x30` errors / DC servo timeouts during playback

**Tested-by:** Blake Gordon <blakegordon@berkeley.edu> (pending kernel rebuild)

## References

- WM8962 datasheet: SPKVDD range 2.5 V–5.5 V
- Device: Microsoft Surface RT, postmarketOS, `linux-postmarketos-grate-7.0.1-r0`
- Related prior work: grate-driver/linux#108 (WM8962 supply properties)


digetx:
Technically, the change is correct, SPKVDD should be 5v. But in practice both 5v and 1.8v regulators are defined as always-on in device-tree and driver doesn't care about actual voltages, hence the change shall have no practical effect and shouldn't fix anything.

blakegordon:
Thanks for the feedback.  We can close the P/R since testing reveals no improvement in getting sound from the device.  

(P.S.  Any suggestions on that?  It's unclear to me whether sound is working for other people running postmarket on Surface RT....)

digetx:
You may try to find from where that error originates in the code

